PR #5, opened 2016-08-29
PEX migration carries weights of prefixes/suffixes of groups and individuals

Boomclaw:
This allows prefixes and suffixes of group to migrate from PEX as well as player specifics using the maximum group weight.


lucko:
👍

PR #11, opened 2016-09-09
Fix PermissionsEx false node migration

HeroCC:
PermissionsEx negated/false nodes begin with a '-' not a '!'. This PR will fix the PEX migration process. [Source](https://github.com/PEXPlugins/PermissionsEx/wiki/Basic-Permissions-Setup#permission-nodes-top)


lucko:
Thanks

PR #18, opened 2016-10-01
Changed constructor of User

MakerTim:
User is a abstract class, and to use it in any package (like a StandaloneUserManager) i need to have access from outside to the constructor

User can still be abstrct, to sign hey you souldn't make a instance of me for api users
but could be created for standalone programs (like mine)


lucko:
You shouldn't be able to make anonymous instances of User, so protected works fine there.

Issue #19, opened 2016-10-10
[mysql] Nullpointer on usernames when there are not already in the system


MakerTim:
https://github.com/lucko/LuckPerms/blob/13bccd1d3f1a6129d8b6c49a99f8cf75f06825bd/common/src/main/java/me/lucko/luckperms/storage/methods/SQLDatastore.java#L160
I got a nullpointer while testing the UI
don't know why exactly, but it seems it returns a null instead of a string containing null

PR #21, opened 2016-10-10
FIX bug #19 - NullPointer

MakerTim:
Fixed nullpointer by adding a null check before its called


lucko:
Resolved in 11facab71c0d9854082de5db7f78a0fc78270de6

PR #30, opened 2016-10-19
Fix temp permissions/groups doesn't removed

mishkapp:
It seems that you forgot to remove temp nodes from datastore so when they expire they remain in datastore and you can see them using "listnodes" command, but they will be marked as "expires in X" where X is time elapsed from expiration moment.


lucko:
The latter part of the bug you mentioned is pretty odd, although you're probably right, I didn't account for expiring temporary permissions in #addNodeUnchecked. The other #setNodes methods in PermissionHolder should've prevented that.

Anyway, my commit varies slightly to yours, only in the sense that I removed any other calls to #auditTemporaryPermissions, in favour of always doing it in the repeating task.

Thanks for bringing it to my attention & taking the time to fix it. 👍

Issue #81, opened 2016-12-12
Reconnect to database


Slind14:
Apparently LuckPerms does not reconnect to the database if the connection was lost leaving the server in a non join-able state.
Any chance you could add an auto reconnect feature to the mysql driver? Usually these are provided by the lib and only need to be enabled.

lucko:
Needs logs showing that. I'm using HikariCP to manage MySQL connections - as far as I know, it will continue to attempt to make connections with your server, even when previous attempts have failed.

Slind14:
```
2016-12-11 23:37:29 [WARNING] [LuckPerms] Processing login for SirWill took 2395ms.
2016-12-11 23:37:29 [INFO] Disconnecting SirWill [/178.63.3.246:57575]: §7§l[§b§lL§3§lP§7§l] §cPermissions data could not be loaded. Please contact an administrator.
```
Unfortunately this is all it tells us without adding further debugging. This happens when the database is down and after starting it back up, you still can't connect. But the permission lookup command still works (reading from cache?).

lucko:
What's probably happening is that those alive connections are waiting the fullest extend of their timeout (30 seconds) before being destroyed by the pool. I'm certainly no expert though - I leave all that stuff down to the author of Hikari. 

I'm not really sure that I'm doing (or rather not doing) anything special. I did a bit of searching - there's nothing obvious I can enable to fix this issue.

I guess, just make sure your database servers are stable. This is certainly the first time I've seen this sort of problem.

Slind14:
hmm that's odd, because the only to fix it, is to restart the server so it reestablishes the connection.
Its running in a cluster, if a node drops it takes up to 20 seconds for it sync back in, so that would be below the 30 seconds.

lucko:
Okay - well, best thing to do it just prevent the DB from going offline in the first place. I know you obviously can't help it, but you'll likely have issues with other plugins too. 

I'm using (pretty much) the default Hikari settings.

Slind14:
Its the first time we had this issue. zPerms, prism and all our custom plugins reconnect just fine.

lucko:
^^^ May help.

Default is 30 seconds, which is a bit long. Now, hopefully, when the DB is down, the connection will fail fast, instead of waiting for connections to come back to life.

Slind14:
The prism version we use does use tomcat jdbc. I'll test the version with reduced timeout and let you know my findings.

Slind14:
[...]

PR #88, opened 2016-12-16
help tame mysql connection settings

SecurityInsanity:
this commit further helps tame the mysql connection pool (and really
reconnecting to the pool even when the underlying db goes down) further
from commit: 31a597c

helps in the case of #81 

this sets up `setValidationTimeout`, and `setConnectionTestQuery`
in order to better (more speedily(?)/fastly(?)/<other adective>ly(?))
handle connection timeouts in a timely manner. (although admittedly
this is already handeled well imo with the cache). however, it could
be better :tm:.

in order to do this I did the following things:
  1. Switch to `TimeUnit.SECONDS.toMillis` instead of manually entering
     milliseconds, and having the time in seconds as a comment. just
     makes it more readable imo. if you want me to change it back I can
     I'm not like attached to this or anything.
  2. Perform more common validation timeouts with `setValidationTimeout`
     this allows us to potentially bump back up the connection timeout
     (although I see no harm leaving it where it is), although allows
     us to still "fail-fast" in a way when doing validations.
  3. Use `setConnectionTestQuery` for people who somehow someway
     may not be using a JDBC4 compliant driver.

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/storage/backing/MySQLBacking.java
+++ b/common/src/main/java/me/lucko/luckperms/common/storage/backing/MySQLBacking.java
@@ -32,6 +32,7 @@ import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.concurrent.TimeUnit;
 
 public class MySQLBacking extends SQLBacking {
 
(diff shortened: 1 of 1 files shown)


Slind14:
Interesting, I hope it resolves the issues with connection loss and not reconnecting. :)

lucko:
Thanks :)

flyingmongoose:
So hey guys I'm kind of re-launching this one, but this is actually a familiar problem to me (I was on the SourceBans for Source-Engine development team). We actually built a fall-back method.

See, SourceBans didn't actually use MySQL to check permissions, it periodically downloaded a local flat file synchronizing the permissions and checked against those instead. This prevented the need to check against MySQL every time, Sped up server operations, and prevented disconnect errors.

Just my 2 cents. You don't have to do it this way of course as the method works (most of the time) but I still ran into it a bit today (but it's because my MySQL DB Host was having problems)

PR #235, opened 2017-04-08
Fix accumulation of temporary perms with a world or server context

creator3:
The old permission wasn't being removed if it was specific to a particular world or server.


lucko:
Thanks 👍

PR #423, opened 2017-08-06
Simplified Chinese Wiki Translation

IzzelAliz:
We had helped translate the wiki to simplified chinese.


lucko:
Hi, firstly, thank you for doing this! I'm sure it took a while. :wink:


I've added your version to the Wiki homepage.   
https://github.com/lucko/LuckPerms/wiki#translations   

...however, I don't believe the translated version belongs in the Readme file. (since this is more for contributors, or those looking to read the code.)

Issue #568, opened 2017-11-28
PostgreSQL failure to launch


VergilPrime:
http://paste.thezomg.com/25205/51184474/
org.postgresql.util.PSQLException: ERROR: syntax error at or near "MODIFY"
  Position: 31

Potential fix: https://github.com/lucko/LuckPerms/pull/569

lucko:
Thanks for reporting - sorry it took a while to get a pushed.

VergilPrime:
No worries!

PR #569, opened 2017-11-28
Potential fix for issue #568

VergilPrime:
I believe this to be a more up-to-date syntax and should be fine with up-to-date SQL servers, but I'm not an expert.


lucko:
Thank you anyway for the contribution :)

I'll try sort another fix.

PR #575, opened 2017-11-30
Fix usernames showing as "MongoDB" #574

### LINKED ISSUE #574: Players usernames are always "MongoDB" after migrating
Migrated from PermissionsEX. We use MongoDB as storage. If a player has not been online since migration and they are looked up by UUID then their username will show as "MongoDB".

We are running with BungeeCord. All servers have `online-mode=false` and `bungeecord: true`. BungeeCord has ip-forwarding enabled.

**Steps to reproduce:**

1. Setup a simple pex structure. Add a few groups and a few users to those groups.
2. Migrate to LP. `lp migration PermissionsEX`. Use MongoDB as backend storage.
3. Run info on a user who had been online before the migration but has not been online after. Look them up via UUID. `lp user <uuid> info`. Their username will then show as MongoDB.

ekrankopio de 2017-11-30 11-09-04
ekrankopio de 2017-11-30 11-14-09

**Edit 1:** Added version data:

ekrankopio de 2017-11-30 11-18-56

Foorack:
Fixes #574

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/storage/dao/mongodb/MongoDao.java
+++ b/common/src/main/java/me/lucko/luckperms/common/storage/dao/mongodb/MongoDao.java
@@ -304,6 +304,7 @@ public class MongoDao extends AbstractDao {
                 if (cursor.hasNext()) {
                     // User exists, let's load.
                     Document d = cursor.next();
+                    String name = d.getString("name");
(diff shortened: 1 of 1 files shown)


lucko:
This was resolved in 003ddc51def7ba232407ceedc3d605433918e567

Thanks for PRing the fix. :)

PR #679, opened 2018-01-12
Cleanup comments

frambinni:
I've noticed a lot of line breaks in the middle of sentences within the comments. I've gone ahead and removed all the ones that don't make sense, but kept line breaks that are there for formatting reasons.

### DIFF
--- a/bukkit/src/main/resources/config.yml
+++ b/bukkit/src/main/resources/config.yml
@@ -243,8 +243,7 @@ vault-debug: false
 # If your MySQL server supports it, the "mariadb" option is preferred over "mysql".
 storage-method: h2
 
-# When using a file-based storage type, LuckPerms can monitor the data files for changes, and then
-# schedule automatic updates when changes are detected.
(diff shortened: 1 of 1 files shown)


spannerman79:
Err you do know some people use nano to edit these config files right?
There is no "word wrapping" for nano making these changes useless

lucko:
These line breaks are intentional and consistent. Each line is wrapped after 100 characters.

Some text editors and viewers (GitHub included...) do not wrap text, which makes it difficult to read the file.



See what I mean? ^^

Thanks for your contribution, but for the reasons outlined above, I won't be merging these changes.

frambinni:
Most text editors have a setting to wrap text, but alright.

lucko:
Perhaps so, but equally as already mentioned, some don't - therefore it makes the most sense to wrap the lines.

Importantly, a large % of the plugins users are likely to be editing the config files using:

* ftp clients
* web based editors in their hosting panels
* command line based programs, such as nano

all of which do not wrap lines.

PR #763, opened 2018-02-18
Let users log in with Nukkit

funniray:
Now users can log in, but support from Nukkit will be required for proper functionallity.


lucko:
This will throw all sorts of errors when the Join event handler tries to retrieve a User object later in the login.

The only fix for this is an async event.

funniray:
Actually that code does work and the permissions are handled properly from the plugin albeit a bit buggy. I was actually able to get the plugin working with this code. The other edits that I also made in this are also needed for the plugin to properly run. The only issues I've found is that /help and the suggested commands don't appear besides that the plugin appears to work properly.

lucko:
Perhaps in your experience it worked - but that code is prone to concurrency issues.

PR #826, opened 2018-03-13
Adding "display-banner" configuration option.

xCykrix:
Since it wasn't possible. I figured I would make it possible and see if you would implement it. If you want anything changed feel free to let me know.

Small change, but it seemed to me like some people would want to use it.

### DIFF
--- a/bukkit/src/main/resources/config.yml
+++ b/bukkit/src/main/resources/config.yml
@@ -52,6 +52,9 @@ use-server-properties-name: false
 # 16 characters or less.
 allow-invalid-usernames: false
 
+# If the startup text should be displayed in console
+display-banner: true
+
 # If LuckPerms should produce extra logging output when it handles logins.
 # Useful if you're having issues with UUID forwarding or data not being loaded.
 debug-logins: false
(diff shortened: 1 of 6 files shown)


lucko:
It's important that the banner is printed before anything else is loaded. The information it contains is very useful for us when providing support / helping users diagnose issues. If, for example, the configuration cannot be loaded (invalid syntax, or whatever), then that information will not be printed.

Additionally, I can't see a reason why it needs a toggle - the message doesn't cause any harm. I've tried to make the logging as concise as possible, whilst still containing all of the necessary information.

You're of course welcome to keep this behaviour in a fork, but I don't think it has a place upstream. :)

xCykrix:
I know the information can be needed. However the bulk of the banner was more what I dislike. The version and etc are indeed useful. However the configuration can be loaded first without and harm as it seems.

Perhaps have a flag to disable just the art over the entire banner? I did oversight that part of the version haha.

Worst go, I'll probably just keep this on my local build server if you have no way to implement this.

lucko:
> However the configuration can be loaded first without and harm as it seems.

> If, for example, the configuration cannot be loaded (invalid syntax, or whatever), then that information will not be printed.


As I said before:

> You're of course welcome to keep this behaviour in a fork, but I don't think it has a place upstream. :)

PR #871, opened 2018-03-27
Fixed command arguments for Nukkit

### LINKED ISSUE #870: Quotes are ignored with Prefixes (on Nukkit?)
Server Version: NukkitX Protocol 1.2.11 (Commit 8f52c901474a019118b849e217c34f034e2c0564)
LuckPerms Version: 4.1.43 (Nukkit)

When I try to do `/lp user <user> meta addprefix <weight> "&fTest Prefix &7"` LuckPerms interperets it as `prefix="&fTest" server="Prefix" world="&7"`. I am using normal quote marks and not the special ones from iOS. I haven't tested this version on Java edition to see if this is a Nukkit-only issue. I'm working on an updated version of EssentialsNK that adds LuckPerms prefix support but I'm waiting on this issue to be resolved before I release it.

funniray:
Fixes #870 
It turns out that quotes are already taken into account by Nukkit.


lucko:
Thanks :)

PR #939, opened 2018-04-25
Added explicit column names in INSERT queries.

D3lph1:
The absence of an explicit indication of the column names leads to some problems. In particular, this happens when using ORM, in which it is inconvenient to specify the order of the columns.


lucko:
LuckPerms doesn't use an ORM?

What issue does this resolve?

D3lph1:
I updated comment.

D3lph1:
Also, do not you think that your code uses different approaches? Why do you list the column names in the other insert insert queries, but not in these?

lucko:
Because the SQL server can't infer the column names on the other queries - they have an `id` auto increment key which is excluded from the data being inserted and populated automatically.

I'm not trying to be intentionally difficult or hostile - I just want to establish if this is fixing a bug or not.

lucko:
Ok - yes I agree. :)

PR #995, opened 2018-05-12
Fix string permission default value

Creeperface01:


Creeperface01:
It was probably already fixed before. Sorry

PR #1162, opened 2018-08-17
Use HIGH priority for login listeners

sgdc3:
Fixes compatibility with auth plugins


lucko:
We're listening on LOW to allow auth plugins to change things at LOWEST, but still provide permissions data for plugins listening on NORMAL.

sgdc3:
@lucko ok that makes sense, but isn't better to listen at HIGH in this scenario?

lucko:
No, because then permissions wouldn't be loaded at NORMAL, when most other plugins are listening.

sgdc3:
That also makes sense xD

PR #1171, opened 2018-08-22
Add Nemisys Support

NycuRO:
Nemisys is a Mini-Bungee for Nukkit.


lucko:
If I'm honest, I don't think this belongs in the upstream project just yet.

The support burden on me to maintain it doesn't really outweigh the number of people likely to be using the platform.

Sorry if that's disappointing for you, but it's why I ask that people open issues before working on substantial PRs. :)

NycuRO:
Actually Nemisys is used by much devs who want create a Network with nukkit and other people. I really like this plugin and i want use them on Nemisys, not per server because i can use much much more ram and performance. 
Nemisys is a good proxy so then.... i don't know. I really appreciate your effort you doing on this and i really love you, but you will do a good job so far if you will want help me with that. It's the same code for nukkit, but adaptated to nemisys. I swear, i don't change anything, you can see on changes page.

NycuRO:
Okey, thank you.

PR #1210, opened 2018-09-09
Use neutral color to display potentially colored names (Fixes #1170)

### LINKED ISSUE #1170: small chat-style "bug"
With the feature:

> group-name-rewrite:
>  moderator: '&2Moderator'

the group-color uverwrite the first character color:

Maybe the plugin can add a "&r" at the end of add renamed group-name for a better usability. ^^

Phoenix616:
This fixes issues with the colors leaking from a group name into messages after the name if it's colored and doesn't end with a reset. Coloring both, the first and the last parentheses makes it look nicer and easier to see if the display name was colored or not (white is mostly associated with not colored at all)

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/model/Group.java
+++ b/common/src/main/java/me/lucko/luckperms/common/model/Group.java
@@ -94,7 +94,7 @@ public class Group extends PermissionHolder implements Identifiable<String> {
     @Override
     public String getFriendlyName() {
         Optional<String> dn = getDisplayName();
-        return dn.map(s -> this.name + " (" + s + ")").orElse(this.name);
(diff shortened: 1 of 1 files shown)


lucko:
Thanks! I'll look over & merge this tomorrow. :)

I want to double check that the #getFriendlyName method modified here isn't being used by the API at all.

Phoenix616:
I took a quick look at it and all I could find (for the group) were messages that get send to the user. The only thing I found where it wasn't directly user facing messages was some logging in the Vault adapter? But these where for players and not groups. :man_shrugging:

lucko:
Thank you :)

PR #1213, opened 2018-09-11
Fixes #1191

### LINKED ISSUE #1191: FastAsyncWorldEdit throws a NPE when LuckPerms is enabled [Nukkit]
Only when LuckPerms is installed, FAWE throws an NPE when you try to do /wea.
image
https://github.com/boy0001/FastAsyncWorldedit/issues/1108

boy0001:
It's valid for null to be passed, it's not a FAWE bug...
https://github.com/NukkitX/Nukkit/blob/master/src/main/java/cn/nukkit/Player.java#L505


lucko:
Thank you :)

PR #1262, opened 2018-10-09
Few optimisations

MrMicky-FR:
Just few optimisations :)

- Remove unnecessary null checks before `instanceof `(`instanceof `always return false when it's null)
- Replace `list.stream().collect(Collectors.joining(" "))` with `String.join(" ", list)`
- Replace `optional.isPresent()` with `optional.ifPresent()`

### DIFF
--- a/bukkit/src/main/java/me/lucko/luckperms/bukkit/migration/MigrationBPermissions.java
+++ b/bukkit/src/main/java/me/lucko/luckperms/bukkit/migration/MigrationBPermissions.java
@@ -116,10 +116,10 @@ public class MigrationBPermissions extends SubCommand<Object> {
                 return CommandResult.FAILURE;
             }
             AtomicInteger userLoadCount = new AtomicInteger(0);
-            users.forEach(s -> {
-                world.loadOne(s, CalculableType.USER);
+            for (String user : users) {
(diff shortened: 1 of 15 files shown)


lucko:
Thanks for your interest in contributing!

I'm aware of most of these - they're IntelliJ inspections! However, I prefer a different style in some cases to what it recommends.

> - Remove unnecessary null checks before `instanceof `(`instanceof `always return false when it's null)

I prefer the explicit check - it makes it clear to anyone reading the node that a null value is expected / accounted for.

> - Replace `list.stream().collect(Collectors.joining(" "))` with `String.join(" ", list)`

I agree with this change.

> - Replace `optional.isPresent()` with `optional.ifPresent()`

I prefer to use an if statement as opposed to using a function in most cases. I know it's more verbose, but it saves on the creation of a Consumer, and is more similar to the `if (x != null) {}` check.

Other things:

* Replacing `.forEach(...)` with a for loop is fine, I generally prefer to do that.
* Removing the explicit `<UUID, LPSubject>` type params is ok as well - they were left in by mistake.

If you could make those changes, I'll happily merge. :)

MrMicky-FR:
Thanks, I made the changes, should be good now :)

lucko:
Thank you :)

Issue #1283, opened 2018-11-05
Stack trace thrown to console on erroneous filter usage


AACJ:
Instead of printing a stack trace, only send the user friendly error message detailing the problem with the syntax.

https://github.com/lucko/LuckPerms/blob/42c8cb1354917b55d9fa00725279cc38c61b1285/common/src/main/java/me/lucko/luckperms/common/commands/misc/VerboseCommand.java#L82

lucko:
Could you make a PR for this? ;)

PR #1284, opened 2018-11-05
Unnecessary stack trace print

AACJ:
Stack traces should be reserved for critical errors within the plugin requiring attention from the developers.

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/commands/misc/VerboseCommand.java
+++ b/common/src/main/java/me/lucko/luckperms/common/commands/misc/VerboseCommand.java
@@ -79,7 +79,6 @@ public class VerboseCommand extends SingleCommand {
             try {
                 parsedFilter = VerboseFilter.parse(filter);
             } catch (InvalidFilterException e) {
-                e.printStackTrace();
                 Message.VERBOSE_INVALID_FILTER.send(sender, filter);
                 return CommandResult.FAILURE;
             }


lucko:
From #1283:
> Instead of printing a stack trace, only send the user friendly error message detailing the problem with the syntax.

So we could possibly pass on the value of `Throwable#getCause` or the result of `#getMessage`.

The point of having the stack trace there is that it at least logs to the console more detailed info about why the filter could not be parsed.

AACJ:
That is one option, or implementing a Logging Level configuration option which dictates whether user error stack traces should be printed or ignored.

lucko:
I don't see why you'd ever want to totally suppress them. 

I'm merely suggesting that if we totally remove any trace of the underlying error from the console, it might be a good idea to include a brief description of the issue in the message sent to the user, as you originally suggested.

e.g.
```diff
--- common/src/main/java/me/lucko/luckperms/common/locale/message/Message.java
+++ common/src/main/java/me/lucko/luckperms/common/locale/message/Message.java
@@ -128,7 +128,7 @@
     /*
      * Commands
      */
-    VERBOSE_INVALID_FILTER("&4{}&c is not a valid verbose filter.", true),
+    VERBOSE_INVALID_FILTER("&4{}&c is not a valid verbose filter. &7({})", true),
     VERBOSE_ON("&bVerbose logging &aenabled &bfor checks matching &aANY&b.", true),
     VERBOSE_ON_QUERY("&bVerbose logging &aenabled &bfor checks matching &a{}&b.", true),
     VERBOSE_OFF("&bVerbose logging &cdisabled&b.", true),
Index: common/src/main/java/me/lucko/luckperms/common/commands/misc/VerboseCommand.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- common/src/main/java/me/lucko/luckperms/common/commands/misc/VerboseCommand.java
+++ common/src/main/java/me/lucko/luckperms/common/commands/misc/VerboseCommand.java
@@ -79,8 +79,7 @@
             try {
                 parsedFilter = VerboseFilter.parse(filter);
             } catch (InvalidFilterException e) {
-                e.printStackTrace();
-                Message.VERBOSE_INVALID_FILTER.send(sender, filter);
+                Message.VERBOSE_INVALID_FILTER.send(sender, filter, e.getCause().getMessage());
                 return CommandResult.FAILURE;
             }
 
```

AACJ:
Agreed. 
Please, however, consider Logging Levels for future implementations as it may be possible to forcibly output stack traces on a higher Logging Level for a larger amount of user input errors for debugging purposes.

BrainStone:
Luck is limited to the levels provided by the platform. And considering the plugin supports many different platforms, there isn’t really an consensus on anything below the info level. Even for the higher ones there are discrepancies.

PR #1400, opened 2019-01-26
show github heading for mit license

untuned:
just a cosmetic change, shows above the file


lucko:
The issue is that there's also a [`HEADER.txt`](https://github.com/lucko/LuckPerms/blob/master/HEADER.txt) file which contains license text.

Even though this is the same underlying licence, the code used by GitHub to detect licenses sees this as being ambiguous. 

I did think about reporting it here: https://github.com/benbalter/licensee

lucko:
A ha - just seen your screenshot. I had the wrong impression then. 

Thanks very much :)

PR #1440, opened 2019-02-21
Default MySQL SSL Settings.

73CHFR34K:
Follow MySQL standarts by define to use SSL or not!

### DIFF
--- a/bukkit/src/main/resources/config.yml
+++ b/bukkit/src/main/resources/config.yml
@@ -136,7 +136,9 @@ data:
     properties:
       useUnicode: true
       characterEncoding: utf8
-
+      useSSL: false
+      verifyServerCertificate: false
+      
   # The prefix for all LuckPerms SQL tables.
   # - Change this is you want to use different tables for different servers.
   table_prefix: 'luckperms_'
--- a/bungee/src/main/resources/config.yml
+++ b/bungee/src/main/resources/config.yml
@@ -133,7 +133,9 @@ data:
     properties:
       useUnicode: true
       characterEncoding: utf8
-
+      useSSL: false
+      verifyServerCertificate: false
+      
   # The prefix for all LuckPerms SQL tables.
   # - Change this is you want to use different tables for different servers.
   table_prefix: 'luckperms_'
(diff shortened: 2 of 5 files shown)


73CHFR34K:
This should stop for new or users with 0 knowledge this error and stop wasting time to search.

```
[Thu, 21. Feb 2019 12:12:47 CET WARN] Thu Feb 21 12:12:47 CET 2019 WARN: Establishing SSL connection without server's identity verification is not recommended. According to MySQL 5.5.45+, 5.6.26+ and 5.7.6+ requirements SSL connection must be established by default if explicit option isn't set. For compliance with existing applications not using SSL the verifyServerCertificate property is set to 'false'. You need either to explicitly disable SSL by setting useSSL=false, or set useSSL=true and provide truststore for server certificate verification.
```

lucko:
This makes it insecure by default? Doesn't seem very sensible to me.

it says right there in the warning message:

> Establishing SSL connection without server's identity verification is not recommended.

73CHFR34K:
No it says too you have to disable it and you can re-enable it by change the config. I only see you are think it makes it insecure but you are wrong with this way of thinking.

Andre601:
I also don't see a reason, why the config would have `useSSL` with a default value of false.
There are only a very few people that report issues with SSL and adding the config option manually isn't that much of an issue.
Most MySQL systems always have SSL active by default, since in most cases they are hosted by a Server-host.
And since those databases are often used by multiple people with multiple different IPs a SSL connection is a must have to make sure, that a secure connection is given.

Andre601:
> I only see you are think it makes it insecure but you are wrong with this way of thinking.

Also as I stated just now, do most people run their server on a host.
And in those cases is the MySQL database most likely on a separate server with separate IP.
Having SSL-check disabled by default would give a potential for people, to abuse that.

73CHFR34K:
do what you need. i dont understand you it is stupid dont use simple settings.

BrainStone:
The only sensible option would be to explicitly enable SSL and certificate verification.

The inexperience of some shouldn’t be the reason for insecurity of many.  
Because if we did that it would be like saying we can’t use passwords anymore because some people forget them.

Insecure by default is a very bad idea. And a much bigger issue than some people struggling setting up the plugin.  
I also believe that if you can’t find the solution to a problem which solution is explained in the official docs in an easy to find spot then you’re not fit to run a server.

lucko:
Going to close for the reasons stated above - I'd rather just keep the default settings.

PR #1537, opened 2019-05-20
v5

lucko:
Still very much a WIP.

I'm still deciding how I want to handle the transition from v4 to v5 - this will most likely break all plugins that integrate with LPs native API. I'm considering supporting both on different branches for a little while.


Andre601:
Maybe have an update with the v4 api and announce the breaking changes before updating at a later date?

Also what are the changes in general?

BrainStone:
I have two suggestions for v5:

- Change the `server` context on the bungee/velocity version to `proxy` and the `world` to `server`. This will simplfy working with per server groups. (If you want I will turn this into a full blown suggestion issue with my full reasoning and all)
- Add a new messaging-service: `auto` and use it as the default. It will behave like `none` behaves right now (do nothing if no SQL storage is being used and behave like `sql` when it is). `none` should always do nothing. And let's get rid of `nosql`.

Andre601:
I don't quite see an actually good reason for the first idea, so explaining it in more detail and especially why it would be useful would be great.
From what I understand is the reason why context is `server` and `world` on both proxy and server-level simplicity.
Instead of having a separate context thing you have to remember, you can just use and apply the same context in the same way: Server is entire server/proxy, world is a specific world/server.

Again I personally would like some more explanation why this should be changed like that.

BrainStone:
Alright. The explanation is pretty straight forward:  
*Per server ranks/permissions/meta*

With the current system to assign a group that works per server on the backend servers *and* the proxies you either have to assign the group twice with different contexts, which is terrible for consitency and maintenace, or use a helper group, which inherits the same group twice for different contexts.

So to simplify this I'm suggesting that `server` reffers to the same thing: The backend server the player is on.  
This make assigning per server things in a network super easy. And you don't have to worry about mainting separate contexts and other shenanigans.

I think it's clear why I'm suggesting this for v5. But to make sure we're all on the same page, this would be a breaking change. So a major update is the right place for that.

Andre601:
I can see what the issues are and can agree that in this case a seperate proxy context for the proxy would make sense.

MrIvanPlays:
finally more simplified api

Jerimu:
I've a rather strange request for version 5...
A track system for only prefixes. So if you join like a Creativeplots server with Tiers from I Till X people can get promoted but only the prefix wil be update. So example the default prefix and track starts with I and i promote the prefix so it will be II and again III and on and on... You get the point.

Tobi406:
Isn't that already possible with the current system, by using two tracks instead of one?

My idea would be the following, I'm just describing it with the groups `one`, `two` and `three`, but I hope you understand me:

# Tracks
- `prefixes_tiers`
-- `prefixes_one` -> `prefixes_two` -> `prefixes_three`
- `tiers`
-- `one` -> `two` -> `three`

# Groups
- `prefixes_one`
-- With permissions:
-- `prefix.1.&7one`
- `prefixes_two`
-- `prefix.2.&7two`
- `prefixes_three`
-- `prefix.3.&7three`

- `one`
-- `group.prefixes_one`
- `two`
-- `group.one`
-- `group.prefixes_two`
- `three`
-- `group.two`
-- `group.prefixes_three`

Then by setting up prefix stacking it should be possible to have only have the prefixes changed by promoting on `prefixes_tiers` and promoting "normal" including the prefixes on `tiers`

# Testing
I am currently not able to test this out, but I'll leave a response here (editing this comment), telling if I made it or not

- **Current state:** Not tested out

Andre601:
Wouldn't it be a good idea to rephrase those flag-names?
They sound a bit weird and Imo overcomplicated.

Ideas of new query param names:
- `INCLUDE_NODES_SET_WITHOUT_SERVER` -> `INCLUDE_NODES_WITHOUT_SERVER`
- `INCLUDE_NODES_SET_WITHOUT_WORLD` -> `INCLUDE_NODES_WITHOUT_WORLD`
- `APPLY_PARENTS_SET_WITHOUT_SERVER` -> `APPLY_PARENTS_WITHOUT_SERVER`
- `APPLY_PARENTS_SET_WITHOUT_WORLD` -> `APPLY_PARENTS_WITHOUT_WORLD`

tl;dr remove the `SET` part since it seems a bit redundant and makes the flags look a bit weirder.
Just a suggestion tho.

lucko:
> Wouldn't it be a good idea to rephrase those flag-names?

Point taken - I've changed them, although they've now ended up even longer 😅

lucko:
I think I'm nearing completion with the API side of things.

Breaking changes to the plugin will come later, my focus currently is getting the API to a state where I'm happy with it, and yeah, nearly there.

I'd welcome any **API related** suggestions anyone has at this stage.

This will probably be our very last chance to make changes.

lucko:
I've hidden the non-API related comments for now, they probably belong in new issues. :))

Andre601:
> Point taken - I've changed them, although they've now ended up even longer 😅

Much clearer that way 👌

EOT3000:
> I'm still deciding how I want to handle the transition from v4 to v5 - this will most likely break all plugins
Maybe just deprecate the old api?

BrainStone:
I'd keep the old API in the plugin jar (and make it functional by delegating the calls to the new API) and of course have it deprecated .  
I'd keep that for at least 5.0, maybe even 5.1.

lucko:
It's quite tricky, there is a lot of stuff that would require proxying.

It's a lot of work for me to try and support this.

BrainStone:
What alternative is there?
Breaking it completely will make server owners an plugin authors equally mad.

Andre601:
I thought we discussed this once where we came to the idea of making an update on the resource pages and inform about the future change of the API to V5 which will break most if not all plugins using the LuckPerms API and that the devs should update their plugins asap.

Andre601:
I also want to bring that up again, because I think it got forgotton or overlooked over the commits made.

> Wouldn't it be a thing to rename `getBytebin()` and `bytebin` to `getBytebinClient()` and `bytebinClient` respectively for consistency?

lucko:
Not super important Andre :p

BrainStone:
> 
> 
> I thought we discussed this once where we came to the idea of making an update on the resource pages and inform about the future change of the API to V5 which will break most if not all plugins using the LuckPerms API and that the devs should update their plugins asap.

The main issue with this approach is that plugins can't support both versions of the API. Or at least not easily. And this is a burden I don't think that should be put in the hands of plugins devs, even if it is possible.

Andre601:
I don't see a reason as to why we should keep this tho...
LuckPerms is not like spigot where people still use older versions (or to be more exact: Where they **shouldn't** use older versions)
LP is compatible with all major versions of Spigot and people should update it anyways so keeping support for an old, worse API version shouldn't be encouraged.
If LP updates then people should update for the simple reason of security and performance improvements... And the same goes for the devs

BrainStone:
As long you will take the support for both angry server owners complaining that the new update broke their other plugins (and therefore they will stay on v4) and the plugin devs why the API was broken completely so they now need to deal with angry server owners themselves, etc.

At least make the read-only part of the API working. That'll satisfy at least 90% of public plugins.

Andre601:
We can't satisfy everyone or else there wouldn't be people on the Discord or in the issues requesting stuff like GUI integration or web editor changes.
When we announce that a breaking change to the API will be launched soon (and provide a way to already use it), then we can already cover a big amount of devs and owners with this...
The people that don't care and won't update basically didn't do that anyway in the past.

Again: I don't see an actual reason why v4 should be kept, even as a read-only thing.
This is not Spigot where it makes sense that 1.8.8 still exists (even tho it shouldn't). Staying with a newer version of LP is almost a requirement now because of previously mentioned reasons (bug fixes and other improvements)
I seriously can't see why there should be any remaining thing for the v4 API once this version here is launched... If you have any actually good reason then please be so nice and elaborate it.

BrainStone:
You tell that to my 2 year old still perfectly fine working plugin.  
I hope you also are aware that devs sometimes abandon plugins. Meaning they never will get updated. But they still work. And then LP throwing backwards compatibility in the trash will make everyone happy because now the server admins have a broken plugin (with possibly no reasonable alternative) that would work if the LP API wouldn't have been broken.

And yes. I do think LP should accept the fact that is has become standard. And therefore needs to be fairly conservative when breaking the API.  
The 4.0-4.4 updates did a great job of that. A plugin written for API 4.0 still works flawlessly with version 4.4. And sure. While full backwards compatibility might not be possible without a massive effort some basic backwards compatbility is pretty much expected.

On the same note, everything that ends up no longer being supported should throw a nice exception that essentially tells the server admin to ask the plugin dev to update to the latest version of the API.

Andre601:
It's funny how you explain PermissionsEx in a nutshell here... Are (not anymore) abandoned plugin that still works fine... Until stuff breaks.
And what will people do when that happens? They search for an alternative. In most cases when a plugin gets abandoned can you expect someone else to take this plugin and update it as a fork.
This essentially happened with [LWC](https://www.spigotmc.org/resources/lwc-extended.69551/), [LightAPI](https://www.spigotmc.org/resources/lightapi-fork.48247/) and not to forget [Essentials](https://www.spigotmc.org/resources/essentialsx.9089/) to mention a few.

An abandoned plugin shouldn't be supported and used anymore and LuckPerms shouldn't encourage this through its API or we have another case of PEX-like people complaining that it works "just fine" despite being god knows how old...

But after all, is this Luck's decision.
And I'm off to bed now.

BrainStone:
I'm gonna stand ground though. I won't be taking any of the "my stuff isn't working anymore" support cases. And there will be tons.

And adding backwards compatibility hardly ever is a bad thing. Especially for popular things and when doing so is easy.

MalzSmith:
I might be a bit late, but it would be nice to have node related events trigger on servers after a network sync

Currently adding a node to a user on a bungee server and then doing a network sync does not trigger the event on the client servers.

Andre601:
About which event, in particular, are you talking?

MalzSmith:
> 
> 
> About which event, in particular, are you talking?

Events in the [me.lucko.luckperms.api.event.node](https://github.com/lucko/LuckPerms/tree/master/api/src/main/java/me/lucko/luckperms/api/event/node) package. I tried with the NodeAddEvent.

lucko:
https://github.com/LuckPerms/extension-legacy-api

Bridge from API v4 --> v5, so existing plugins coded against older versions of LuckPerms will continue to work.

Currently supports everything (read/write etc), with the one big exception of event listeners - that will probably come later though.

Andre601:
How exactly would this jar be added?
Would it be added into an extension folder or similar? Or would it be just put into the plugins folder?

lucko:
Into an extension folder. Reason being: I don't want to clutter the main plugin with legacy compatibility code.

Andre601:
I assume it's now ready for release?

PR #1661, opened 2019-08-06
Update config description of messaging service

Andre601:
The LuckPerms configs contain a description of the messaging services you can use.
Those are partially confusing and maybe even misleading.

For example does the description of `none` state that it...
> Disables the service.

...while `sql` says the following information:
> Uses the SQL database to form a queue system for communication. Will only work when 'storage-method' is set to MySQL or MariaDB. **This is chosen by default if the option is set to 'none' and SQL storage is in use. Set to 'notsql' to disable this.**

This PR adds a bit more description about how `none` works in certain situations and also lists and explains the `notsql` option since according to the description of sql is this a valid and available option which shouldn't be kept secret.


lucko:
:+1:

PR #1677, opened 2019-08-13
Add console note patch

Andre601:
This PR adds a note for the player to check the console if they're a server admin and have access to the console.
Many server owners encounter those database errors and while it is a clear and logical way to check the console when this happens, do many not realize to do exactly that.
This added message should now improve this.


Andre601:
Closed because of useless and already implemented PRs

PR #2060, opened 2020-03-15
Resolve server crash from concurrent permissions request

PikaMug:
User of the Quests plugin reported issue https://github.com/PikaMug/Quests/issues/1121 with thread dump https://hastebin.com/dujucemeno.shell

Attempting to iterate players' effective permissions concurrently results in a server crash. To wit:

https://github.com/lucko/LuckPerms/blob/b84518e1f1a95cc49b790b5ba9a9a5242f661489/bukkit/src/main/java/me/lucko/luckperms/bukkit/inject/permissible/LuckPermsPermissible.java#L213-L217

While the unmodifiable `getPermissionMap()` offers protection, it is my understanding that the map implementations (namely HashMap.entrySet) are non-volatile fields and are lazily initialized such that access to these fields is unsafe from all other threads. See https://stackoverflow.com/a/29058755

https://github.com/lucko/LuckPerms/blob/01e17adf82d00b81b4b5e86f0ce115d64811bcdc/common/src/main/java/me/lucko/luckperms/common/cacheddata/type/PermissionCache.java#L74

Therefore, I believe the issue can be addressed by returning a new instance of `this.permissions`. I've forgone using a basic HashMap as the class already imports ConcurrentHashMap. Thanks for your time!

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/cacheddata/type/PermissionCache.java
+++ b/common/src/main/java/me/lucko/luckperms/common/cacheddata/type/PermissionCache.java
@@ -71,7 +71,7 @@ public class PermissionCache implements CachedPermissionData {
     public PermissionCache(QueryOptions queryOptions, CacheMetadata metadata, CalculatorFactory calculatorFactory) {
         this.queryOptions = queryOptions;
         this.permissions = new ConcurrentHashMap<>();
-        this.permissionsUnmodifiable = Collections.unmodifiableMap(this.permissions);
+        this.permissionsUnmodifiable = Collections.unmodifiableMap(new ConcurrentHashMap<>(this.permissions));
 
         this.calculator = calculatorFactory.build(queryOptions, metadata);
         this.calculator.setSourcePermissions(this.permissions); // Initial setup.


lucko:
As you can see, `this.permissions` is already a `ConcurrentHashMap`, so I'm not sure what creating a copy would do to fix the problem?

https://github.com/lucko/LuckPerms/blob/01e17adf82d00b81b4b5e86f0ce115d64811bcdc/common/src/main/java/me/lucko/luckperms/common/cacheddata/type/PermissionCache.java#L73

Iterating over `ConcurrentHashMap.entrySet()` is thread safe.

About the issue itself, I'll reply here: #2062

lucko:
Seems the underlying issue isn't being caused by this - so I'm going to close. Thank you for attempting a fix though! :)

PikaMug:
@lucko Yes, but iterating over `unmodifiableMap().entrySet()` is _not_ currently thread safe:

https://github.com/lucko/LuckPerms/blob/01e17adf82d00b81b4b5e86f0ce115d64811bcdc/common/src/main/java/me/lucko/luckperms/common/cacheddata/type/PermissionCache.java#L103-L105

We should create a new instance because, right now, it's merely wrapping the original collection. Here's a somewhat-similar example: https://stackoverflow.com/questions/19726337/concurrentmodificationexception-in-unmodifiable-collection

I might restate that this issue results in a server _crash_, not an exception, so it's my understanding this really shouldn't be left for other plugins to pick up on.

lucko:
> We should create a new instance because, right now, it's merely wrapping the original collection. Here's a somewhat-similar example

It's not a similar example, because it is wrapping a `HashMap`, not a `ConcurrentHashMap`.

> I might restate that this issue results in a server crash, not an exception, so it's my understanding this really shouldn't be left for other plugins to pick up on.

As explained on the other issue, I don't think this actually *is* an issue, and also don't think it was the cause of the crash.

PikaMug:
I wrote _somewhat-_similar, recognizing that.

What do you suspect the cause is then?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

lucko:
electronicboy's suggestion on the Quests issue seems the most plausible to me.

PikaMug:
I'm assuming you mean https://github.com/PikaMug/Quests/issues/1121#issuecomment-599201817 which, to me at least, doesn't read as a suggestion and more of an analysis of the timing report. Anyway, given that we have three conversations on the topic going at once, maybe the issue is starting to lose focus.

It has been alluded to that this may ultimately be considered an exploit. Assuming I've been correct in my assessment, it would only take a server running LuckPerms + any plugin(s) which can call `LuckPermsPermissible#getEffectivePermissions` in a frequent manner, such as during a Bukkit event, to crash a server. At this point, I must stress that this PR potentially has a lot to gain, but very little to lose. If I'm wrong, the worst thing this PR would do is a maybe lose a few milliseconds on performance.

lucko:
> If I'm wrong, the worst thing this PR would do is a maybe lose a few milliseconds on performance.

Not quite, there are other issues with the behaviour of the code. `permissionsUnmodifiable` is a view of the `permissions` map - copying in the constructor means this is a copy, not a view, and therefore will not update automatically.

I glossed over that because the fix is unnecessary anyway.

I've slightly improved the performance of `getEffectivePermissions` in this commit 05b82cdbabd17c3565b094297344e4c0097fe36c - maybe that will help some what.

The root cause though if likely that a client was spamming those packets illegally.

PikaMug:
Yes, it's a copy, but as far as I can tell the method is provided for external use; not internal, so it makes little to no difference in practice, unless you're expecting someone to call the method ad infinitum, in which case we'd run into a parallel issue.

I appreciate the commit, as it seems Guava will ultimately create a copy of the entry set, likely to avoid this very problem:
https://github.com/google/guava/blob/559e1ac1359e8d5dc0941aa72691698185d65407/guava/src/com/google/common/collect/ImmutableMap.java#L382

If by "illegally" you mean via hacked client, I'm not certain that was the case, but even so this was a welcome conclusion. Thanks for your time!

PR #2091, opened 2020-03-23
Support SporkBukkit's modifications to the permissions API

Electroid:
There are some performance forks of Bukkit, such as [SportPaper](github.com/Electroid/SportPaper), that rely on patch that modifies the `PermissibleBase`.

```patch
 public class PermissibleBase implements Permissible {
     private ServerOperator opable = null;
     private Permissible parent = this;
-    private final List<PermissionAttachment> attachments = new LinkedList<PermissionAttachment>();
-    private final Map<String, PermissionAttachmentInfo> permissions = new HashMap<String, PermissionAttachmentInfo>();
+    private final Set<PermissionAttachment> attachments = new LinkedHashSet<PermissionAttachment>();
+    private final ListMultimap<String, PermissionAttachmentInfo> permissions = ArrayListMultimap.create();
```

This small change breaks an underlying assumption in `PermissibleInjector` and `LuckPermsPermisslble`: that the `attachments` field is a `List`. Thus, all logins are rejected with the following error:

```log
[LuckPerms] Exception thrown when setting up permissions for <uuid> - <username> - denying login.
[WARN]: java.lang.ClassCastException: java.util.LinkedHashSet cannot be cast to java.util.List
[WARN]:        at me.lucko.luckperms.bukkit.inject.permissible.PermissibleInjector.inject(PermissibleInjector.java:103)
[WARN]:        at me.lucko.luckperms.bukkit.listeners.BukkitConnectionListener.onPlayerLogin(BukkitConnectionListener.java:193)
```

I've proposed a very simple fix that mitigates this issue so more servers can use your plugin. Thanks and happy to answer any questions if you have them. Cheers!

### DIFF
--- a/bukkit/src/main/java/me/lucko/luckperms/bukkit/inject/permissible/LuckPermsPermissible.java
+++ b/bukkit/src/main/java/me/lucko/luckperms/bukkit/inject/permissible/LuckPermsPermissible.java
@@ -56,6 +56,7 @@ import java.util.ListIterator;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.Spliterator;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 
@@ -351,10 +352,12 @@ public class LuckPermsPermissible extends PermissibleBase {
      *
      * Some (clever/dumb??) plugins attempt to add/remove/query attachments using reflection.
      *
+     * Some bukkit forks change the attachments list to be a set, so support both interfaces.
+     *
      * An instance of this map is injected into the super instance so these plugins continue
      * to work with LuckPerms.
      */
-    private final class FakeAttachmentList implements List<PermissionAttachment> {
+    private final class FakeAttachmentList implements List<PermissionAttachment>, Set<PermissionAttachment> {
 
         @Override
         public boolean add(PermissionAttachment attachment) {
@@ -397,6 +400,11 @@ public class LuckPermsPermissible extends PermissibleBase {
         @Override
         public Iterator<PermissionAttachment> iterator() {
             return ImmutableList.<PermissionAttachment>copyOf(LuckPermsPermissible.this.hookedAttachments).iterator();
+        }
+
+        @Override
+        public Spliterator<PermissionAttachment> spliterator() {
+            return ImmutableList.<PermissionAttachment>copyOf(LuckPermsPermissible.this.hookedAttachments).spliterator();
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


lucko:
This looks ok, but my understanding is that a little more work is required to properly support SportBukkit.

https://github.com/Electroid/SportPaper/blob/master/patches/api/0024-Improve-Permissible-interface.patch

These extra methods would need to be implemented too - I'm not sure how keen I am to do this in LuckPerms.

Andre601:
The question I myself have now is, how does this patch improve the Permissible interface?
And considering that a majority of people use Spigot and PaperMC (which LP supports) and I myself never heard of SportPaper, is the question if it is really worth it to support a fork, that literally breaks plugin compatibility in a way.

Electroid:
The fix is only intended to support *basic* functionality. Yes, there is some more logic that would have to be added for full support, but I would agree you don't want to diverge your project for a fork. I also think this change is focused enough so that it shouldn't have any harm on the future maintainability of your code.

lucko:
> I also think this change is focused enough so that it shouldn't have any harm on the future maintainability of your code.

I agree!

> The fix is only intended to support basic functionality. Yes, there is some more logic that would have to be added for full support, but I would agree you don't want to diverge your project for a fork.

Sure, it might make it load correctly, but methods such as `PermissionAttachmentInfo getEffectivePermission(String name)` are going to cause a runtime exception when called - this arguably would count as "basic functionality", which was the reason for my comment before.

There are also some other parts of the code which are silently failing:

https://github.com/lucko/LuckPerms/blob/0729be0f0a6ff9097ede6d060dbfbc56f9567df4/bukkit/src/main/java/me/lucko/luckperms/bukkit/inject/dummy/DummyPermissibleBase.java#L81-L102

(admittedly those exceptions should probably be rethrown! A poor choice to ignore them on my part)

joandotcodina:
(https://github.com/lucko/LuckPerms/issues/2141) This branch solved my issue.
Ty @Electroid

joandotcodina:
This pull, solves the problem that user can't loggin with SportPaper versions...

lucko:
Just as an update, the reason this hasn't been merged is because it doesn't actually add full support for SportBukkit - as noted above more work is necessary.

Two ways this can move forward:

1. More changes made to LuckPerms to support SportBukkit - would probably mean LuckPerms would need to depend on SportBukkit's API instead of Spigot and implement the extra methods it adds.

2. SportBukkit removes the patches which modify the permissions API to make itself compatible with permission plugins which expect "unmodified" CraftBukkit permissions implementation classes.

Approach no 2 seems the most sensible to me, especially since these patches don't really add a whole lot in the way of functionality to plugins, and since most plugins write against Bukkit/Spigot anyway.

However, a well formed PR implementing no 1 will still be accepted (whether it this one or another).

Electroid:
Thanks for entertaining this change, I'm going to go with the route of fixing this in SportPaper.

lucko:
Great 😄

Electroid:
For those coming from Google search, this is now fixed: https://github.com/Electroid/SportPaper/commit/3d0a5ccef48dee073c3fdbd960eefc54149eb002

lucko:
Awesome, thank you!

PR #2135, opened 2020-04-03
Fix the link to Storage types comparison

caramat:
The link to [Storage types](https://github.com/lucko/LuckPerms/wiki/Storage-types) seemed to be outdated


Andre601:
Would it be possible, that you could also update the information about the sql messaging type?
It still mentions that LP will default to it when you use `none` while using a MySQL-like database... This has changed to my knowledge as `none` was changed to `auto`

@lucko please correct me here if I'm wrong.

lucko:
Thanks @caramat!

@Andre601, fe139f8420091bba479421ed66788dd2fd468e79

Issue #2203, opened 2020-04-20
MySQL: SSL warning despite useSSL=false


Picajoluna:
Hello,

I'm trying to set up LP with a **locally** hosted MySQL db.
Everything works -- the permissions data is correctly stored in the database.
I seem to be running into a persistent/spammy (and somewhat annoying) console warning. This warning is flashed ~10 consecutive times at boot, and every 10-20 minutes:

```
[18:15:51 WARN]: Mon Apr 20 18:15:51 UTC 2020 WARN: Establishing SSL connection without 
server's identity verification is not recommended. According to MySQL 5.5.45+, 5.6.26+ and 5.7.6+ 
requirements SSL connection must be established by default if explicit option isn't set. For 
compliance with existing applications not using SSL the verifyServerCertificate property is set to 
'false'. You need either to explicitly disable SSL by setting useSSL=false, or set useSSL=true and 
provide truststore for server certificate verification.
```

To access the db, I created a mysql user with an associated password, and granted it all privileges.

Here's my config file, where I have in fact set `useSSL=false`:
```yml
137     # By default, the following options are set to enable utf8 encoding. (you may need to remove
138     # these if you are using PostgreSQL)
139     #   useUnicode: true
140     #   characterEncoding: utf8
141     #
142     # You can also use this section to disable SSL connections, by setting:
143     useSSL: false
144     verifyServerCertificate: false
145     properties:
146       useUnicode: true
147       characterEncoding: utf8
```

When I [look up](https://stackoverflow.com/questions/34189756/warning-about-ssl-connection-when-connecting-to-mysql-database) the issue, it seems that using `useSSL=false` should fix the issue, yet it has not for me this far. Anyone know what the problem might be? Thanks :pray:

benwoo1110:
it should be:
```
    properties:
      useUnicode: true
      characterEncoding: utf8
      useSSL: false
      verifyServerCertificate: false```

as stated in the wiki here: https://github.com/lucko/LuckPerms/wiki/Storage-system-errors#mysql-ssl-errors

Picajoluna:
Argh my bad. Thanks so much!

PR #2211, opened 2020-04-22
Improve "properties" section of config files

Andre601:
This PR improves the properties section (`data.pool-settings.properties`) by moving the mentioned `useSSL: false` and `verifyServerCertificate: false` parts below the `characterEncoding: utf8` part, acting as uncommented options.

The goal of this is to prevent confusion, as some people will assume, that you just need to uncomment the mentioned options and not add them below the properties one (see #2203 for a recent case)

### DIFF
--- a/bukkit/src/main/resources/config.yml
+++ b/bukkit/src/main/resources/config.yml
@@ -139,12 +139,13 @@ data:
     #   useUnicode: true
     #   characterEncoding: utf8
     #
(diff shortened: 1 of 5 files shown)


lucko:
Thanks, I've included a similar change in 07b2df7c7e2ee141bdc30ceb96e505abba3a98a2

Andre601:
Alright. Just wanted to mention that I saw your commit xD

Issue #2299, opened 2020-05-13
Request: Adding multiple servers in Context counting as this OR that server


zettlinger:
I have multiple servers and as i understood from a Mod on Discord (Larry) i would have to add each node to each server either by duplicating data or adding in specific groups.
This can get pretty messy and complicated , i am working on a 9 server setup to add to the 4 i have already. Even the group would mean i can only add the exact same list of perms OR i need to make a ton of groups again.

What i would like is that i can add multiple server: server 1 , server 2
The editor allows this but apparently reads it as that that node wont get active unless the player is on 2 servers at once. this makes no sense.

Please kill this data redundancy and just read SERVER and WORLD context as OR iso AND

ps a simple way to COPY a line would also help not with redundance but ease of adding

ApertureDevelopment:
In a way it makes sense if it's the same context.

You are right that a user can't be on two servers/worlds at the same time, a User can always only be on one world, making "OR" the better choice here, but when its different contexts (e.g. Server and World ) they should be AND connected

So as a not working example:
``(World 1 OR World2) AND (Server1 OR Server2)``

lucko:
Added in ed85ab1bfd1673e10a313179936ad211c313910f :)

PR #2300, opened 2020-05-13
Change ContextSet 'satisfy' behaviour

lucko:
The suggestion was:

> different contexts should be AND connected, but if its the same context, OR makes more sense

So that's what this does.

### Example
User has the following current contexts:
```
server=survival
world=overworld
gamemode=creative
wg-region=spawn
```

Previously, defining a node with contexts `server=survival world=overworld world=nether` would not have applied, as the user would have needed to be in:

> `server=factions` AND `world=overworld` AND `world=nether` (not possible).

However, with this change, the user needs to be in:

> `server=factions` AND (`world=overworld` OR `world=nether`) (possible!).

### Considerations

* Might break existing usages
* Would make us incompatible (technically) with the sponge context spec.

It could be put behind a config option, but my current thoughts are that it's quite a nice change, and we should just change it!

I can't think of a use case where the old behaviour (compared to what is being suggested here) would be useful, but happy to be proven wrong!

Would like to hear peoples thoughts. :))

@bloodmc @zml2008 I figure you might be interested ^^ as creators/heavy users of the contexts system, I'd appreciate any opinions/ideas you have if you'd like to share them 😊

### DIFF
--- a/api/src/main/java/net/luckperms/api/context/ContextSet.java
+++ b/api/src/main/java/net/luckperms/api/context/ContextSet.java
@@ -197,12 +197,28 @@ public interface ContextSet extends Iterable<Context> {
     }
 
     /**
+     * Returns if the {@link ContextSet} contains any of the given context pairings.
+     *
+     * @param key the key to look for
+     * @param values the values to look for
+     * @return true if the set contains any of the pairs
+     * @since 5.1
+     */
+    default boolean containsAny(@NonNull String key, @NonNull Iterable<String> values) {
+        for (String value : values) {
+            if (contains(key, value)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
      * Returns if this {@link ContextSet} is fully "satisfied" by another set.
      *
      * <p>For a context set to "satisfy" another, it must itself contain all of
-     * the context pairings in the other set.</p>
-     *
-     * <p>Mathematically, this method returns true if this set is a <b>subset</b> of the other.</p>
+     * the context keys with values in the other set, with at least one of the values
+     * matching.</p>
      *
      * @param other the other set to check
      * @return true if all entries in this set are also in the other set
--- a/common/src/main/java/me/lucko/luckperms/common/context/contextset/ImmutableContextSetImpl.java
+++ b/common/src/main/java/me/lucko/luckperms/common/context/contextset/ImmutableContextSetImpl.java
@@ -40,6 +40,7 @@ import net.luckperms.api.context.MutableContextSet;
 
 import org.checkerframework.checker.nullness.qual.NonNull;
 
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Objects;
@@ -145,14 +146,11 @@ public final class ImmutableContextSetImpl extends AbstractContextSet implements
         } else if (other.isEmpty()) {
             // this set isn't empty, but the other one is
             return false;
-        } else if (this.size() > other.size()) {
-            // this set has more unique entries than the other set, so there's no way this can be satisfied.
-            return false;
         } else {
             // neither are empty, we need to compare the individual entries
-            Set<Map.Entry<String, String>> entries = this.map.entries();
-            for (Map.Entry<String, String> e : entries) {
-                if (!other.contains(e.getKey(), e.getValue())) {
+            Set<Map.Entry<String, Collection<String>>> entries = this.map.asMap().entrySet();
+            for (Map.Entry<String, Collection<String>> e : entries) {
+                if (!other.containsAny(e.getKey(), e.getValue())) {
                     return false;
                 }
             }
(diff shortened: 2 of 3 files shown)


ApertureDevelopment:
Issue #2299 is related

zml2008:
This is something I've thought a bit. While I'm not in favour of any particular solution just yet, some of the ways I've debated solving this are:
1. Allow the value side of a context to specify a pattern (say, a permission could be set in `world={overworld,nether}`. This would make context parsing more complicated, but opens the door for more extensive patterns. I've done something like this in PEX to support ip range type contexts with CIDR notation.
2. Perform that same parsing/expansion within commands -- so `/lp user whoever set my.permission true world={overworld,nether}` (apologies for the butchered command syntax) would set that permission twice, once for each world context. This would of course lose the information that this one permission is explicitly set to qualify in two worlds, but would (at least for me in PEX) result in the fewest implementation changes.

Even if you choose to go ahead with your proposal, I expect most plugins won't be affected by the minor break in spec -- GD would probably be the most affected but it already uses LP directly so it's not as much of a concern.

Do you have thoughts on these concepts?

lucko:
If I was to write things again from the ground up, `#1` is what I'd favour.

However, representing `#1` in a nice way programatically & in storage isn't super straightforward. I've followed your recent PEX dev stuff a bit, and the solution of requiring "context types" to be registered makes sense, as this registration can then provide the behaviour for parsing command args, saving to/from storage, matching other contexts, etc - however, this is quite a big change and I don't think at this point in time is something I want to do in LP. Just having string pairs has been sufficient for - as far as I can recall, all use cases I've ran into so far - but to be fair I haven't got IP address contexts or anything like that. 😛 

`#2` has been suggested before, and is an option - however, the benefits of having less entries "in the file" and also when viewing users/groups permissions is lost as everything gets reduplicated.

I support a (kind of) glob style expansion in permission nodes, such as `mines.teleport.{1-4}` - however, the expansion is applied when permissions are resolved at runtime, not when the command is ran. (in other words, `mines.teleport.{1-4}` is actually what ends up "in the file")

This could be done for contexts too I suppose, (would be a mixture of `#1` and `#2` in some ways) - but having in code/the API, a `String` `"world"` mapped to another `String` `"overworld,nether"` just doesn't seem very nice.

The reason for considering this most recent idea (the one I describe above) is because it seems to:

* solve the problem of wanting to use OR matching in contexts
* without breaking the vast majority of existing usages
* with minimal code diff
* without needing to change the (effective) `Multimap<String, String>` signature of contexts

I'll sit on this for a bit longer & see if any more ideas come up before making a call.

Thank you zml for the feedback/ideas, I appreciate it 😃

Turbotailz:
Are there any known cases where the AND matching is actually used for duplicate context keys? I know LP does support it currently and would be interesting to know if anyone has found a context that works in this way. This would probably need to be considered as well (config option best solution in that case IMO).

Laarryy:
A (good?) reason not to switch to being able to define many of the same context type in one node- for extracontexts, if you have to be in a region within another region, you'd be unable to set a context to only work in that region. (could be worked around by naming the inner region something unique that other inner regions aren't, but still a concern if people use that function)

I like the option to have it 'locked' behind a config toggle that defaults to false, so people that update or redownload the plugin and start fresh don't have to worry about it, but people who want it can use it.

lucko:
> A (good?) reason not to switch to being able to define many of the same context type in one node- for extracontexts, if you have to be in a region within another region, you'd be unable to set a context to only work in that region. (could be worked around by naming the inner region something unique that other inner regions aren't, but still a concern if people use that function)

Basically yeah, this would prevent you from trying to set something only where regions overlap. So so niche though haha.

Anyway, I've made it configurable. However, it defaults to using the new behaviour, so users won't need to change anything in their configs after updating to get the nicer mode.

lucko:
This will get merged shortly after lucko/LuckPermsWeb#238 is resolved. Code is written, just a case of deploying. (hopefully sometime tomorrow!)

bloodmc:
If I pass contexts `gd_claim=basic` and `gd_claim=admin`, does LP check in a specific order? If so, I could potentially use this as a way to prioritize contexts.

lucko:
> If I pass contexts gd_claim=basic and gd_claim=admin, would LP check in a specific order? If so, I could potentially use this as a way to prioritize contexts.

This PR doesn't change that behaviour, no. Fundamentally context sets are still unordered.

> Now one question...

Pretty much, yes! With this change, you could set a permission with

```
server=survival
gd_claim_default=basic
source=minecraft:player
target=minecraft:creeper
target=minecraft:zombie
```

and the permission would be matched by passing contexts with **either** `target=minecraft:creeper` or `target=minecraft:zombie`.

> Would LP know to store this with the first permission

It's not currently doing the cleanup automatically, although this could be added. :)

> If so, I can see this as being a very valuable way of cleaning permission storage.

Yep, that's the idea!

> I also am not fond of having config options when it comes to context functionality. The main reason being if I do plan on using this functionality , I would have to tell each server to enable it. It would be nice if we could come to a general consensus without breaking current plugin compatibility. Currently, I do not see how this would break current plugins but I'm not 100% positive.

Understood, I'm pretty much all for enabling it by default (this is what I've done so far). As with lots of other things in LP though, I've buried a config option deep in the file just in case an experienced admin wants to change it.

bloodmc:
Sounds great. Thanks!

bloodmc:
Sounds great. 

> It's not currently doing the cleanup automatically, although this could be added. :)

GD servers would greatly appreciate it if added =) Without having to reapply all permissions.

lucko:
:tada:

PR #2334, opened 2020-05-26
Increase world column max size

### LINKED ISSUE #2298: world column too small in database
The `world` column is only VARCHAR(36) but that is too small for my use case.

A plugin I am using ([World System Plugin](https://www.spigotmc.org/resources/worldsystem-●-the-one-world-per-player-solution-●-gui-●-highly-configurable-●-1-8-1-15.49756/)) uses the players' uuid plus an id for the world name which ends up longer than 36 letters.

I wanted to use the context system in luckperms to give players perms in only their world but I can't because of this error:
```
13.05 12:55:07 [Server] WARN Caused by: org.h2.jdbc.JdbcBatchUpdateException: Value too long for column """WORLD"" VARCHAR(36) NOT NULL": "'id1-6e05911c-d04a-4cba-ab04-fe0bf50c8388' (40)"; SQL statement:
13.05 12:55:07 [Server] WARN INSERT INTO `luckperms_user_permissions` (uuid, permission, value, server, world, expiry, contexts) VALUES(?, ?, ?, ?, ?, ?, ?) [22001-199]
```

pixelsquared:
fixes #2298

### DIFF
--- a/common/src/main/resources/me/lucko/luckperms/schema/h2.sql
+++ b/common/src/main/resources/me/lucko/luckperms/schema/h2.sql
@@ -6,7 +6,7 @@ CREATE TABLE `{prefix}user_permissions` (
   `permission` VARCHAR(200)       NOT NULL,
   `value`      BOOL               NOT NULL,
   `server`     VARCHAR(36)        NOT NULL,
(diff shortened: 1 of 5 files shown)


lucko:
Thanks, this will do for now :))

PR #2432, opened 2020-07-04
Add option to use bytebin for exports/imports

Turbotailz:
Allows users to export to bytebin instead of saving to file.

- `/lp export --upload` - returns the 10 digit bytebin code
- `/lp import <code> --upload`

image

A couple things to note:

- It does not perform any checks on the import to see if the JSON is formed correctly, I'm not sure if this is handled by the Importer class already.
- There is no way to copy the code, which may be annoying to some users. I considered letting users click a link for the bytebin upload but IMO it seems unnecessary.

Let me know if there is anything that should be changed. I'm not very comfortable using Java so there will likely be things that can be done better :P


lucko:
Looks good to me, thanks!

PR #2489, opened 2020-07-24
Fix import with no users object

### LINKED ISSUE #2488: Importing an export file with `--without-users` flag throws NPE
On a Sponge server, user tried importing an export file that was exported with the `--without-users` flag and it throws an NPE:
https://pastebin.com/v2TT4W3Y

Haven't tested with Bukkit/etc but I'm guessing it may be the same since this points to /common/ . I found it in the code and told a user to use it, and the import failed. Another solution is to simply export all and import all, and then bulkupdate the users away. However, the flag works on export so it should probably work on import too. It's due to the importer looking for "users" and finding no JSON obj. for it, as far as I can tell

Just thought I should let you know! I'm also going to work on a PR to fix this today, might as well give it a shot.

Laarryy:
This will fix #2488 - I even tested it 😄 

Apologies for the three commits, first was me updating my branch using only GH like a fool, and then the IDE did me a dirty and 'fixed' your imports before I committed and I didn't notice, so i had to go back and undo that 😛 

Also, please let me know if there is any sort of versioning that needs to be done in the PR, and if so where, and I'll be happy to do it! If not, that's a pretty neat system you've got!


lucko:
Thanks, merged in the above commit, but with a slightly different approach to solving the issue.

The versioning is as magic as you think it is ;)

PR #2672, opened 2020-10-15
Changed Discord logo to white in Readme

devsword7:
I think that white on green fits the theme a bit better than blue on green.


lucko:
Thanks - but I'm happy with how it looks at the moment :)

PR #2687, opened 2020-10-24
feat : add server context boolean with group manager migration

fhebuterne:
Hello,

I added second boolean parameter on groupmanager migration command because i need to migrate on LuckPerms but using groups and permissions server context on each server where groupmanager is present.

First boolean already exist to define world context as global or not.
Second boolean do the same but with permissions and groups server context.

/lp migration groupmanager true true -> To migrate as global context
/lp migration groupmanager true false -> To migrate as server context

It work perfectly on my test server.
Sadly, i don't have any time to implement this and test it on other permissions plugin migrations  for now

Thanks


Turbotailz:
One issue I see with this is that we get a lot of users asking for help specifically with the GM migration because they don't understand that they need to add `true` to the end of the command. Adding another argument here will only confuse things a lot more. I think we need to have defaults in this case, where they would both default to `true` and migrate permissions globally.

fhebuterne:
@Turbotailz Hmm, in my case i needed to split server context and world migration, because i wanted to migrate permissions and groups with server context but not using world context so i put in globally, if i put in one argument on the same true, my specific case isn't possible, add information in documentation isn't possible ?

lucko:
Hey, thank you for the PR.

Sorry it's taken me a while to respond. I've been deliberating as to whether this change is a good idea. I think ultimately I've come to the conclusion that adding more flags to the migration command is a mistake. (I regret adding the one for `migrateAsGlobal`! It seems to confuse people)

Thank you for contributing and submitting your improvement as a pull request, but I think at this time it doesn't make sense to merge into the main project.

Hopefully you understand where I'm coming from! :blush:

Issue #2696, opened 2020-10-31
A new command (Quality of life?)


UwuItsDavid:
### Description

**I would like a command such as /lp set caps, to automatically capitalize the first letter of every group created. As of now, I have to go and use /lp group <rankname> setdisplayname <name> for each group.

the setdisplayname you gotta do it manually for each group
./lp set caps can automatically do it for all created groups
/lp group owner setdisplayname Owner

It's a name, so it needs to be capitalized ;-;
but lp defaults to lowercase
So it's a quality of life feature, and would be easier on those who are creating 10+ ranks.**

### Proposed behaviour

The idea is that when groups are created, even when using a capital in the name, it always goes to lowercase. For players using placeholders to display rank names, it's an annoyance to have the rank name be lowercase.
So lets say a player creates 15 ranks, (Gamemode prison has lots of ranks, 26+). As of now, he has to go and use the command **/lp group owner setdisplayname Owner**, and 14 more times for the other ranks to make them appear Capitalized. If this suggestion is added, once said user creates all of his ranks, he won't have to touch **/lp group <rank> setdisplayname <name>**, he can type **/lp set caps** which will capitalize the first letter in every group, as the displayname. I'm not expecting it to make the actual group in the files/databases capitalizes, just a mass "displayname" command.

Fefo6644:
I reckon that a good solution to this would be an enhancement to the `creategroup` command, it being something along the lines of `/lp creategroup <name> [weight] [displayname]` (which I might or might not be working on already :d).

UwuItsDavid:
> I reckon that a good, more flexible solution to this would be an enhancement to the `creategroup` command, it being something along the lines of `/lp creategroup <name> [weight] [displayname]` (which I might or might not be working on already :d).

That would work as well! :) Can't wait to see if you do, or do not add it!

PR #2698, opened 2020-11-02
Allow creategroup command to take weight and display name as optional arguments

Fefo6644:
An idea I've been having for a while (mainly for the weight), and a possible solution for #2696 \:)


lucko:
Looks good, thanks!

PR #2699, opened 2020-11-02
Fix bulk update SQL stats lookup bug

Fefo6644:
Fixes a bug for the bulk update command failing on SQL based storages when tracking operation stats with multiple constraints, making the "impact lookup" query `... WHERE field1 = 'value1'field2 = 'value2'field3 = 'value3' ...`


lucko:
Thanks!

PR #2732, opened 2020-11-18
Fix translations .verbose.enabled and .enabled-recording arguments

Fefo6644:


Fefo6644:
This would need to be changed in the crowdin project too.... yikes

PR #2809, opened 2021-01-04
Update mongodb driver to latest version and save uuid with the new standard (version 4)

SamBolduc:
The mongodb java driver save uuid with the old standard (version 3), now, I force it to save uuid with the new standard (version 4)

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -51,6 +51,6 @@ dependencies {
     }
     compile 'com.zaxxer:HikariCP:3.4.5'
     compile 'redis.clients:jedis:3.3.0'
-    compile 'org.mongodb:mongo-java-driver:3.12.2'
+    compile group: 'org.mongodb', name: 'mongodb-driver-legacy', version: '4.1.1'
     compile 'org.yaml:snakeyaml:1.23'
 }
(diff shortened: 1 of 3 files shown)


lucko:
Would this change not make the storage incompatible with any existing data saved in a different format?

SamBolduc:
> Thank you for the PR!
> 
> A question though:
> Would this change not make the storage incompatible with any existing data saved in a different format?

Yes, this will break existing ids. It is not possible to transform a uuidv3 to a uuidv4 but an id saved in v3 is unusable since it does not correspond to mojang user's id.

It would be necessary to ensure that version 4 is used only if no data has yet been recorded. Another alternative would be to add an option in the configuration file allowing you to choose the type of id you want.

lucko:
I'm not sure about that - I'm pretty sure the current serialization is working ok - perhaps it would be possible to automatically migrate them to a new format

SamBolduc:
There is the stored _id with uuid saved as v3
image

and there is the stored _id with uuid saved as v4
image

There is my real uuid from mojang:
image

As you can see, the uuid saved as v3 doesn't match the mojang uuid.  This makes it not possible to perform database searches using mojang uuid

lucko:
Thanks for the extra info, I also did a bit of reading about the various UUID types Mongo uses.

A seamless transition is really important here: ideally, we assume that users might not even update all of their servers at the same time, so older versions of LP should be able to continue to function correctly.

So yes, a seamless migration from "current" to "recommended" UUID encoding is important. Otherwise, I'm inclined to just leave things as they are - it does seem to work somewhat ok afterall!

PR #2847, opened 2021-01-24
Substitute unnecessary mutability for immutability

Conclure:


lucko:
Hi, thanks for the PR - I appreciate it!

However, I'm not entirely sure how this is an improvement? The 'mutability' wasn't causing a problem before, and this change makes the class more complex by skipping out on the automatic error handling provided by the static method used previously.

Furthermore, when reloading, we kinda want to discard the existing `YamlConfiguration` & create a new one instead of merging into the existing!

Issue #1365, opened 2019-01-01
RCON


FarrisBris:
When I execute Luckperms commands thru RCON it returns a blank line.

Chappy202:
What version of luckperms are you running?

FarrisBris:
4.3.2

Chappy202:
Try upgrading to 4.3.17 and see if it fixes your issue

lucko:
It's an issue with plugins which handle commands async / attempt to send responses to command senders at a later time.

FarrisBris:
> 
> 
> It's an issue with plugins which handle commands async / attempt to send responses to command senders at a later time.

As of my understanding, there is no way at all that RCON will work with LuckPerms?

lucko:
At the moment, yes that is the case.

FarrisBris:
> 
> 
> At the moment, yes that is the case.
> 
> If Spigot merged the previously mentioned pull request, then it would work, but I can't influence that process.

Well that is a huge rip. Thank you.

lucko:
Yep indeed it is a shame. :/

brandinarsenault:
@lucko,

Sorry to comment on this old issue thread. I'm assuming this pull request has not been merged and there is no update posted? Unfortunately, I cannot review the pull request using that link as my login for their Jira does not grant me access.

Thank you for all your work on this, it is a fantastic plugin and project!

Brandin.

lucko:
Yep, pull request is still open, but hasn't received any comment / update in over a year.

straight-shoota:
This is really a bummer. I suppose the fix hasn't landed yet still.

I'd like to issue lp commands over RCON to initialize groups and memberships from a web UI. The commands itself execute, so that's technically fine there's no feedback and I can't verify the existing settings.
Are there any suggestions how to achieve that instead? I could directly access the database, but I figured letting LuckPerms handle that would avoid interference and potential consistency issues. But looks like this might be the better solution when RCON doesn't work properly.

maximov-ru:
Pleease reopen this bug.

brandinarsenault:
> Pleease reopen this bug.

This isn't a LuckPerms bug - so theres no reason to have this open. It's on Spigot to fix, not Lucko.

macbrayne:
Weirdly enough the Fabric version doesn't work either with RCON, this shouldn't be due to some Spigot limitations, right?

HyperPaint:
[...]

PR #2861, opened 2021-01-27
fix(lucko#1365): run bukkit RCON commands synchronously

cbarber:
bukkit/spigot have an outstanding issue where async messaging does not function over
RCON. this works around the issue until a fix PR is accepted

### DIFF
--- a/bukkit/src/main/java/me/lucko/luckperms/bukkit/BukkitCommandExecutor.java
+++ b/bukkit/src/main/java/me/lucko/luckperms/bukkit/BukkitCommandExecutor.java
@@ -27,6 +27,7 @@ package me.lucko.luckperms.bukkit;
 
 import me.lucko.luckperms.bukkit.util.CommandMapUtil;
 import me.lucko.luckperms.common.command.CommandManager;
+import me.lucko.luckperms.common.command.CommandResult;
 import me.lucko.luckperms.common.command.utils.ArgumentTokenizer;
 import me.lucko.luckperms.common.config.ConfigKeys;
 import me.lucko.luckperms.common.sender.Sender;
@@ -36,6 +37,7 @@ import org.bukkit.command.Command;
 import org.bukkit.command.CommandSender;
 import org.bukkit.command.ConsoleCommandSender;
 import org.bukkit.command.PluginCommand;
+import org.bukkit.command.RemoteConsoleCommandSender;
 import org.bukkit.command.TabExecutor;
 import org.bukkit.entity.Player;
 import org.bukkit.event.EventHandler;
@@ -45,6 +47,8 @@ import org.checkerframework.checker.nullness.qual.NonNull;
 
 import java.util.List;
 import java.util.ListIterator;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 public class BukkitCommandExecutor extends CommandManager implements TabExecutor, Listener {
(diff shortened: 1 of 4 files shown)


Laarryy:
This looks phenomenal! fixing #1365 has been impossible for a while, but I think that it needs a config option (people hate synchronous operations that they're unaware of) so that people who enable it are aware and willing to use a sync operation regardless of potential downsides

cbarber:
`(sender instanceof RemoteConsoleCommandSender)` limits synchronous operations to just RCON. User commands and console commands will operate asynchronously. There is also WARNING before each response for RCON users that this command could block the game thread. I feel like a feature flag for a bug fix with no breaking changes is an anti pattern.

Laarryy:
The warning imo is less preferred than a toggle - if you want to use it, having a potential flood of warnings would not be preferable, while if you don't want to use it, rcon commands should be blocked entirely. This isn't exactly a bugfix as it does not correct the source problem, it rather works around the problem by forcing RCON commands to run synchronously. As this is, as I see it, closer to a feature, it may be something users want to entirely disable, and if they want to use it then
1. Hardcoding a warning in English for a plugin that supports several languages is not preferred
2. Using it often, with knowledge of what it does, would make many users displeased with the flooded logs of the same warning that they know about and chose to use anyways.

This is just how I see it, Luck has the final say 😄

cbarber:
You convinced me that the warning is noise and should be removed. Done.

I'm not convinced this is a feature. The problem is lp commands over RCON do not output a response. A response is the only indication of successful or failure on RCON: https://developer.valvesoftware.com/wiki/Source_RCON_Protocol

I'm also not convinced RCON commands should run asynchronously. I couldn't find an RFC to confirm this, but RCON protocol seems to be serial in nature with one (or more) response packets following a command packet. Making this command asynchronous can lead to race conditions when bulk commands are run sequential.
```
lp creategroup test
lp group test permission set foobar true
lp group test rename test2 
lp group test2 permission set foobaz false
```
The above is contrived, but any bulk set of operations that mutates your permission state should be run in order.

My use case is initializing a docker cluster with all required groups and permissions by piping the commands through RCON. This lets my seed scripts remain agnostic to any database changes, and I just need to watch out on major rev updates.

Fefo6644:
I reckon there should be a setting (something like `rcon-sync-operations` or whatever, false by default) that if false it blocks the operation and warns the sender instead but if true it just doesn't do that and runs the command unsafely; this would be more like `vault-unsafe-lookups` when making operations for offline players on the main server thread.

cbarber:
I'm not clear how synchronous execution of a RCON command is unsafe.

Fefo6644:
The issue isn't the RCON command, the issue is that the operation will effectively run on the main server thread and, if the operation takes long time to complete (e.g. load user data, run a heavy database query, bulk operations on text-based storage methods, etc), it will cause lag; that is <i>the</i> reason why `vault-unsafe-lookups` is false by default and throws a runtime exception.

cbarber:
We have the following requirements:
* RCON should run synchronously for consistency/adhere to the protocol
* `onCommand` must complete quickly to avoid blocking the main game thread

Proposed changes:
* Add support for integerKey
* Include rcon-timeout-seconds defaulted to 5

Commits updated.

Laarryy:
Including a timeout is a great option, but does not solve the main problem - primarily that rcon commands run sync and sync operations are toggleable in LuckPerms. As Fefo brought up, there are even other examples - `vault-unsafe-lookups` des not have a timeout, but rather a toggle. I believe it is better design to allow a toggle because having a default, limited, timeout will not match some users' specifications, while a toggle allows them the freedom to choose entirely. 

I'm thinking here of a user who wants to use the RCON feature and *needs it*, for whatever reason, to prevent all other operations until it is done. Is this smart? Probably not, but it does not mean that the plugin should choose for the user, the choice should be left with the user entirely. Having a timeout option is fantastic, I suggest, however, that it defaults to 0 and that a toggle is also added. That way, a user can choose to enable it without any timeout, or they can enable it and set whatever timeout they want.

cbarber:
The timeout supports that behavior. Set the timeout to Integer.MAX_VALUE. The pedantic among you will argue that is only a 68 year timeout but I'm willing to wait and see if anyone hits it :trollface:

lucko:
Hey, thanks for the PR!

However, I would prefer it if we took a slightly different approach: a boolean config option like `rcon-run-commands-sync` which defaults to false, but when true, uses the blocking behaviour.

LuckPerms commands are not suited to being ran on the main thread - most of them involve some sort of blocking i/o which will make the server (game) thread lag if executed sync.

The reason for defaulting false - rcon is still perfectly functional at the moment for LP commands (without causing any lag), which imo is better than having cmd feedback + lag.

lucko:
If you'd be happy to make those changes, I will merge the PR. :)

If not, I don't mind making the changes myself, since I agree it would be a useful option to have.

lucko:
Please feel free to re-open if/when you implement those changes. :)

Issue #2774, opened 2020-12-15
Changing from H2 to MySQL database causes issues with luckperms


LonlyGamerX:
Recently decide to switch to MySQL due to i think it would be better than mysql. So i setup and enter everything in correctly and check and all of the info is the same i use for my currency plugin which works. However once i switch to MySQL i keep getting this issue in console --- >
```
[01:08:01 ERROR] [LuckPerms]: Exception occurred whilst loading data for 3be4767d-a9a9-47f1-b104-327f68467865 - LonlyGamerX
java.util.concurrent.CompletionException: java.sql.SQLSyntaxErrorException: Table 'luckpermsdata.luckperms_players' doesn't exist
        at me.lucko.luckperms.common.storage.Storage.lambda$future$0(Storage.java:92) ~[?:?]
        at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1700) ~[?:?]
        at net.md_5.bungee.scheduler.BungeeTask.run(BungeeTask.java:66) ~[travertine-163.jar:git:Travertine-Bootstrap:1.16-R0.4-SNAPSHOT:fe5acbf:163]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) ~[?:?]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) ~[?:?]
        at java.lang.Thread.run(Thread.java:834) [?:?]
Caused by: java.sql.SQLSyntaxErrorException: Table 'luckpermsdata.luckperms_players' doesn't exist
        at me.lucko.luckperms.lib.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:120) ~[?:?]
        at me.lucko.luckperms.lib.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:97) ~[?:?]
        at me.lucko.luckperms.lib.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122) ~[?:?]
        at me.lucko.luckperms.lib.mysql.cj.jdbc.ClientPreparedStatement.executeInternal(ClientPreparedStatement.java:953) ~[?:?]
        at me.lucko.luckperms.lib.mysql.cj.jdbc.ClientPreparedStatement.executeQuery(ClientPreparedStatement.java:1003) ~[?:?]
[...]

PR #2864, opened 2021-01-29
Fix millenary issue on SQL storage systems

### LINKED ISSUE #2842: Tables not being created when loading plugin on MySQL database.
### Description

When I start my server I get this error:
`java.util.concurrent.CompletionException: java.sql.SQLSyntaxErrorException: Table 'permscv.luckperms_groups' doesn't exist`
I have the latest version of LP and I have tried making new of everything but it doesn't work. I have figured out that its not actually creating the table in the DataBase so it isn't finding it.

### Reproduction steps

1. Put your database credits in config.yml
2. Restart the server
3. See error
4. If you want to further make sure that its not actually creating the tables then you can go to you MySQL Command Line Client and type `show tables;` and it will show you that only `luckperms_messenger` has been created.

### Environment details

* Server type/version: `PaperMC` running version `1.16.4` build `415`
* LuckPerms version: `v5.2.92`

Full LP startup log: https://bytebin.lucko.me/s39CuTJIoh

Fefo6644:
Fixes #2842, #2774, and a number of duplicate issues where LP would not create the database tables even though it would connect properly (additionally, creating and leaving the `{prefix}messenger` table as the only table on the database); issue is it was scanning through the entirety of the sql server, so if another database on the same server already had the `{prefix}user_permissions` table, it would return true. Fixed by passing the catalog (essentially the db name) to the getTables query as "filter".

Tested on MariaDB and MySQL.

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/storage/implementation/sql/SqlStorage.java
+++ b/common/src/main/java/me/lucko/luckperms/common/storage/implementation/sql/SqlStorage.java
@@ -980,7 +980,7 @@ public class SqlStorage implements StorageImplementation {
     }
 
     private static boolean tableExists(Connection connection, String table) throws SQLException {
-        try (ResultSet rs = connection.getMetaData().getTables(null, null, "%", null)) {
+        try (ResultSet rs = connection.getMetaData().getTables(connection.getCatalog(), null, "%", null)) {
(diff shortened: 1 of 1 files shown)


lucko:
Ahhhh amazing - thank you!

PR #2867, opened 2021-01-30
Improve CraftBukkit detection

Fefo6644:
Lately on the discord server I've noticed a number of CraftBukkit + offline mode server logs that LP wasn't detecting as CraftBukkit (but CB being CB it still [doesn't call APPLE](https://hub.spigotmc.org/jira/browse/SPIGOT-3541)), to then find out recent CB builds don't show the version as `git-Bukkit-<short sha1> ...` but `<some number?>-Bukkit-<short sha1> ...`; this PR fixes that, tested on both old and new CB builds.


lucko:
LGTM, thanks

PR #2874, opened 2021-02-04
RabbitMQ support

ruViolence:
RabbitMQ messaging support


lucko:
This looks really great - thanks so much!

common/src/main/java/me/lucko/luckperms/common/messaging/rabbitmq/RabbitMQMessenger.java
```diff
@@ -0,0 +1,185 @@
...
+                        if (!this.firstStartup) {
+                            this.parent.plugin.getLogger().info("RabbitMQ pubsub connection re-established");
+                        }
+                        return true;
+                    } catch (Exception ignored) {
+                        return false;
+                    }
+                } else {
+                    try {
+                        this.parent.channel = this.parent.connection.createChannel();
```
ruViolence, line 160: @lucko Oops, after that things should be declared like above
```java
String queue = this.parent.channel.queueDeclare("", false, true, true, null).getQueue();
this.parent.channel.exchangeDeclare(EXCHANGE, BuiltinExchangeType.TOPIC, false, true, null);
this.parent.channel.queueBind(queue, EXCHANGE, ROUTING_KEY);
this.parent.channel.basicConsume(queue, true, new ChannelListener(), (consumerTag) -> { });
```

lucko: Ah.

I mean, if the connection isn't "dead" - does it need to be reestablished?

ruViolence: It works like this: `connection -> channel -> listener (exchange/queue/consumer)`.

If the channel or connection was closed, then the exchange queue and consumer (listener) will be closed and must be re-created in both cases, if the connection was closed then new channel must be created.

BomBardyGamer:
You god

PR #2877, opened 2021-02-05
Added virtual host support for RabbitMQ

BomBardyGamer:
Tested on latest BungeeCord (`git:BungeeCord-Bootstrap:1.16-R0.5-SNAPSHOT:3d701fb:1545`), no reason why it wouldn't work on any other system since the configs follow the format I found in them, and this is all in common rather than implementations. I will happily test it on all systems though on request.

Virtual hosts are used to split up things into different things basically (I'm not very good with explanations, see [here](https://www.rabbitmq.com/vhosts.html)), which allows you to isolate all your different projects that use it, so there's no interference.

### DIFF
--- a/bukkit/src/main/resources/config.yml
+++ b/bukkit/src/main/resources/config.yml
@@ -253,6 +253,7 @@ redis:
 rabbitmq:
   enabled: false
   address: localhost
+  vhost: '/'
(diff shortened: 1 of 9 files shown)


lucko:
Looks good, thanks!

PR #2903, opened 2021-02-20
Fix PostgreSQL authentication type 10 is not supported

Hai-tun:


lucko:
Looks great, thanks!

Also fyi I updated the gradle wrapper which should fix the issues you had building :) 7579bbe5496ad0e2e1bfab8f468c708bc4b3e69b

Hai-tun:
Thank you very much. After my test, gradle can be built normally. I am very sorry to be so anxious to get your answer, but your discord group does not welcome me. Maybe I don’t understand your customs, but @somebody is not the most basic Notify? @lucko (This is a friendly @

PR #2906, opened 2021-02-22
Fix holder type and name on hover and click events in meta info entries

Fefo6644:
Inherited meta nodes would show "Click to remove ... from \<holder\>" where \<holder\> would be the holder the command was run on, not the original meta node holder. This would often cause confusion when trying to delete the clicked node when inherited.

### Before

image

### After

image


lucko:
LGTM, thank you!

PR #2937, opened 2021-03-09
Safer way of getting GeneratedEventClass package

KR33PY:
When using Mohist 1.16 (and maybe other Forge+Bukkit hybrids), GeneratedEventClass.class.getPackage() is null for whatever reason.
This tiny change implements another way of getting the package and restores
compatibility with Mohist & Co.


lucko:
Thanks - although I should also say: really it would be sensible to actually find the root cause of this problem instead of patching usages of methods which should be working.

Chances are there will be other plugins (silently) failing because of this, which is obviously not good!

KR33PY:
> Thanks - although I should also say: really it would be sensible to actually find the root cause of this problem instead of patching usages of methods which should be working.

I totally agree, however [Oracle javadoc](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getPackage--) states that `getPackage` is nullable.

lucko:
Yes, but there is no good reason for the information to be intentionally excluded, so I imagine it is a bug in Mohist / whatever.

PR #2958, opened 2021-03-24
Fixes #2957

### LINKED ISSUE #2957: 'meta settemp' ignores the provided temporary modifier
### Description

Trying to replace/accumulate a temp meta node on top of an existing one results in command failure because "Player already has temp meta".

### Reproduction steps

1. Set a temporary meta node
2. Try setting it again on the same holder with `accumulate` as temporary modifier
3. It no work

### Expected behaviour

I would expect it to:
* If using `accumulate` to add the time to the existing node
* If using `replace` to, well, replace the expiry time.

### Environment details

* Server type/version: `Paper` running version `1.16.5` build `567`
* LuckPerms version: `v5.3.14`

image

### Any other relevant details

The command checks for existing nodes already (temp or not) before adding the new node, effectively ignoring the temporary modifier.
https://github.com/lucko/LuckPerms/blob/2e62c98c4564683e5ba7af367173413739e89dcb/common/src/main/java/me/lucko/luckperms/common/commands/generic/meta/MetaSetTemp.java#L82-L92

Fefo6644:
This does not check for existing nodes, but it first removes existing temp meta nodes for the same key and different values and then applies the node w/ the temporary modifier.

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/commands/generic/meta/MetaSetTemp.java
+++ b/common/src/main/java/me/lucko/luckperms/common/commands/generic/meta/MetaSetTemp.java
@@ -45,10 +45,10 @@ import me.lucko.luckperms.common.sender.Sender;
 import me.lucko.luckperms.common.util.Predicates;
 
 import net.luckperms.api.context.MutableContextSet;
+import net.luckperms.api.model.data.DataMutateResult;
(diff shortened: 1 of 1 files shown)


lucko:
Looks good, thank you!

PR #3051, opened 2021-06-04
Updated to Gradle 7

BomBardyGamer:
Does what it says in the title. I've also update shadow to go with this, and also switched every `compile` to `implementation` (since `compile` has been removed in 7), and switched to using `filter` to replace tokens in resource files, rather than the previously used `expand`.


lucko:
Thanks for this - I will merge once I have a spare moment to ensure Jenkins is up-to-date and configure a different Gradle install

lucko:
> switched to using filter to replace tokens in resource files, rather than the previously used expand.

Why was this, out of interest? @BomBardyGamer

lucko:
Other than that, looks good, Jenkins is ready for Java 16 + Gradle 7 so I can merge now 😄

bukkit-legacy/build.gradle
```diff
@@ -1,15 +1,15 @@
...
-    compile 'com.google.code.gson:gson:2.7'
-    compile 'com.google.guava:guava:19.0'
+    implementation project(':bukkit')
+    implementation 'com.google.code.gson:gson:2.7'
+    implementation 'com.google.guava:guava:19.0'
 }
 
 shadowJar {
-    archiveName = 'luckperms-bukkitlegacy.jarinjar'
+    archiveName 'luckperms-bukkitlegacy.jarinjar'
```
emilyy-dev, line 12: Seeing that the gradle scripts are being reworked, this should be `archiveFileName` (same for the other cases); `get/setArchiveName` is deprecated

emilyy-dev (changes requested):

lucko:
Thanks :)

PR #3069, opened 2021-06-14
Implement enable-ops and auto-op on Fabric

emilyy-dev:
Also Luck please suggest a better name for `FabricOtherListeners` :joy: could not think of anything better at 2 a.m.


lucko (approved):

lucko:
Thanks!

PR #3070, opened 2021-06-15
General API improvements

emilyy-dev:
Add `NodeMap::addAll(Iterable<Node>)` and `NodeMap::addAll(NodeMap)`
Add `PermissionHolder::getInheritedGroups(QueryOptions, boolean)`
Initial commit for BulkUpdate API (BulkOperation? Aiming at renaming that sector of the plugin, at least for the API :stuck_out_tongue:)
Add `PluginMetadata::getConfigDirectory()`, aiming to be used in extensions

### DIFF
--- a/api/src/main/java/net/luckperms/api/model/PermissionHolder.java
+++ b/api/src/main/java/net/luckperms/api/model/PermissionHolder.java
@@ -300,6 +300,24 @@ public interface PermissionHolder {
     @NonNull @Unmodifiable Collection<Group> getInheritedGroups(@NonNull QueryOptions queryOptions);
 
     /**
+     * Gets a collection of the {@link Group}s this holder inherits nodes from.
+     *
+     * <p>This method exists to avoid the verbosity of converting an existing {@link QueryOptions}
+     * to its builder form solely to change the {@link Flag#RESOLVE_INHERITANCE} flag.</p>
+     *
+     * <p>This calls {@link #getInheritedGroups(QueryOptions)} with the modified query options.
+     * See the description for it for the specific behavior of this method.</p>
+     *
+     * @param queryOptions       the query options
+     * @param resolveInheritance asd
+     * @return a collection of the holder's direct parent groups
+     * @since 5.4
+     */
+    default @NonNull @Unmodifiable Collection<Group> getInheritedGroups(@NonNull QueryOptions queryOptions, boolean resolveInheritance) {
+        return getInheritedGroups(queryOptions.toBuilder().flag(Flag.RESOLVE_INHERITANCE, resolveInheritance).build());
+    }
+
+    /**
      * Removes any temporary permissions that have expired.
      *
      * <p>This method is called periodically by the platform, so it is only necessary to run
--- a/api/src/main/java/net/luckperms/api/model/PermissionHolder.java
+++ b/api/src/main/java/net/luckperms/api/model/PermissionHolder.java
@@ -303,15 +303,17 @@ public interface PermissionHolder {
      * Gets a collection of the {@link Group}s this holder inherits nodes from.
      *
      * <p>This method exists to avoid the verbosity of converting an existing {@link QueryOptions}
-     * to its builder form solely to change the {@link Flag#RESOLVE_INHERITANCE} flag.</p>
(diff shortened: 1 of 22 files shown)


lucko:
Happy to accept some of these changes, but I feel that they would be better off in separate PRs. There's some things here which look incomplete, and some changes which I'm not so sure on..

If you'd like to pursue this further please let me know and I can give some more specific feedback, but for now I'll close this - thanks :)

api/src/main/java/net/luckperms/api/model/PermissionHolder.java
```diff
@@ -299,6 +299,26 @@
...
+     * See the description for it for the specific behavior of this method.</p>
+     *
+     * @param queryOptions       the query options
+     * @param resolveInheritance whether the resulting collection should include inherited groups
+     *                           beyond the holder's direct parent groups
+     * @return a collection of the groups the holder inherits from
+     * @since 5.4
+     * @see #getInheritedGroups(QueryOptions)
+     */
+    default @NonNull @Unmodifiable Collection<Group> getInheritedGroups(@NonNull QueryOptions queryOptions, boolean resolveInheritance) {
```
lucko, line 318: `getInheritedGroups` ... without inheritance? That doesn't seem right.

api/src/main/java/net/luckperms/api/model/data/NodeMap.java
```diff
@@ -109,6 +109,25 @@
...
+    /**
+     * Adds the nodes from another node map.
+     *
+     * <p>Effectively calls {@link #addAll(Iterable)} with the result from
+     * {@link #toCollection()}.</p>
+     *
+     * @param nodeMap the node map to merge into this one
+     * @since 5.4
+     */
+    void addAll(@NonNull NodeMap nodeMap);
```
lucko, line 129: I think the first one covers this sufficiently, users wanting to add a `NodeMap` can call `.toCollection()` themselves.

api/src/main/java/net/luckperms/api/platform/PluginMetadata.java
```diff
@@ -46,4 +48,11 @@
      */
     @NonNull String getApiVersion();
 
+    /**
+     * Gets the config directory path
+     *
+     * @return the directory path in where the plugin stores various configuration files
+     */
+    @NonNull Path getConfigDirectory();
```
lucko, line 56: Why? I don't see a good usecase for this.

lucko (commented):
Added some separate comments for a few of the changes... leaving Bulk Update:

* I'm not happy with the overall quality/functionality of the bulk update system as it is. (that of couse has nothing to do with this PR)
  * See https://github.com/lucko/LuckPerms/issues/1836 for a long-long wishlist of desired changes and improvements.
  * In particular, it was never really updated to properly support contexts, and instead targets the `server` and `world` contexts especially.
  * I haven't started on this because it's a lot of work, for a feature which I think is probably only ever used occasionally. Also it's already rather complex.
* I don't think it makes sense to expose the Bulk Update function as an API in this way - the classes in the common module are based on the command layout - not what necessarily makes best sense as API. A fluid-like builder API would probably be better, a bit like what jOOQ does.

... but, until the greater bulk update system is improved, I don't think it would be sensible to provide an API (as it could restrict changes later down the line, if they ever get worked on)

api/src/main/java/net/luckperms/api/platform/PluginMetadata.java
emilyy-dev: My main thought was configurable extensions, it can be made in any way to be fair but this was what first came to mind.

api/src/main/java/net/luckperms/api/model/data/NodeMap.java
emilyy-dev: Fair enough I guess

api/src/main/java/net/luckperms/api/model/PermissionHolder.java
emilyy-dev: Yeah I guess it can be `getParentGroups` only with no extra parameters then; I thought of adding this method initially because I often needed to get the direct parent groups of a user, it does add up and it does yield some ugly looking code.

PR #3121, opened 2021-07-29
Updated Java/Gradle and FabricMC components

DrakoTrogdor:
Mostly just minor tweaks in order to compile it in my environment.
Updated:
    Java 16
    Gradle 7.1.1
    Minecraft 1.17.1 (fabric server)
    FabricMC API 0.37.0+1.17
    yarn 1.17.1+build.14
    Fabric-loader 0.11.6


emilyy-dev:
I don't think that "updating to Java 16" can be considered a "minor tweak", I don't have any stats in hand but I think it's safe to assume that many servers that use LuckPerms don't run on Java 16.

underscore11code:
Looking at the [global bStats statistics for Bukkit](https://bstats.org/global/bukkit), only ~46% of servers are running J16. We can assume that LP probably has a similar percentage.

emilyy-dev:
And that is Bukkit only, there aren't any precise statistics over the other platforms LP runs on (bStats Sponge being opt-in doesn't really contribute to any safe assumptions)

lucko:
Implemented in ba522597a893c6e8aae86420ebeee7c81106d52e & 93683c3a2f55c20bd37c6a7eb28aa088a130f8c8 - although I have opted to keep the other platforms on Java 8 for now.

PR #3129, opened 2021-08-06
Group rename - Save renamed group before syncing with storage

emilyy-dev:
When using the newly introduced `--update-parent-lists` flag in the group rename command, it had the catastrophically undesired side effect of refreshing the new/renamed group from storage, discarding **every** permission that was added after the group creation.

The primary change is really in GroupRename, but to ensure that the bulk operation + syncing happens *after* the group is saved, I ended up adapting the StorageAdapter to return the messaging service request's future for continuation purposes.


lucko:
LGTM, thanks!

PR #3170, opened 2021-09-26
adding note about required operator

RedstoneFuture:
If you deactivate operator via LP, you have side effects with basic vanilla features. I add a small note in the config about that.


lucko:
Thanks!

PR #3198, opened 2021-11-03
Simplify code using Collection::isEmpty

LewUwU:


lucko:
I disagree that it is a simplification in this case - thanks for the PR nonetheless :)

PR #3226, opened 2021-11-29
Add pagination to listgroups command

### LINKED ISSUE #3211: `listgroups` pagination
### Description

Currently, `/lp listgroups` command is not paginated. We have many groups on our server, the list is not readable and is not even completely shown on the chat.

### Proposed Behaviour

`/lp listgroups` should print first page of groups. The usage should be `/lp listgroups [page]`, where `page` defaults to `1`.

### Extra Details

_No response_

Aberdeener:
Closes #3211
Let me know of any style changes you'd like, especially in terms of the text messages!

Screen Shot 2021-11-29 at 4 32 43 PM

Screen Shot 2021-11-29 at 4 32 50 PM

Sorry for the commit mess, forgot to fetch from upstream before making my edits.

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/commands/group/ListGroups.java
+++ b/common/src/main/java/me/lucko/luckperms/common/commands/group/ListGroups.java
@@ -39,6 +39,7 @@ import me.lucko.luckperms.common.util.Predicates;
 
 import java.util.List;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class ListGroups extends SingleCommand {
     public ListGroups() {
@@ -63,7 +64,7 @@ public class ListGroups extends SingleCommand {
                     return i != 0 ? i : o1.getName().compareToIgnoreCase(o2.getName());
                 }).collect(Collectors.toList());
 
-        List<List<Group>> pages = Iterators.divideIterable(groups, 7);
+        List<List<Group>> pages = Iterators.divideIterable(groups, 13);
 
         if (pageIndex < 0 || pageIndex >= pages.size()) {
             page = 1;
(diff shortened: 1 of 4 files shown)


common/src/main/java/me/lucko/luckperms/common/commands/group/ListGroups.java
```diff
@@ -53,15 +55,27 @@ public void execute(LuckPermsPlugin plugin, Sender sender, ArgumentList args, St
...
+        if (pageIndex < 0 || pageIndex >= pages.size()) {
+            page = 1;
+            pageIndex = 0;
+        }
+
+        Message.SEARCH_SHOWING_GROUPS.send(sender, page, pages.size(), groups.size());
+        Message.GROUPS_LIST.send(sender);
+
+        for (Group group : pages.get(pageIndex)) {
+            List<String> tracks = plugin.getTrackManager().getAll().values().stream().filter(t -> t.containsGroup(group)).map(Track::getName).collect(Collectors.toList());
```
montlikadani, line 77: It would be good to cache the `plugin.getTrackManager().getAll().values()` collection outside of this for loop, as the `getAll` implementation creates a copy of the `ImmutableMap` each time it is called. But if you are not paying attention to performance, as this command is called once usually it is not necessary. Just a suggestion.

lucko:
LGTM, can we increase the default list size to whatever will fill a vanilla chat window? I think it's 15 lines? So 15 minus two for the header bit?

common/src/main/java/me/lucko/luckperms/common/commands/group/ListGroups.java
Aberdeener: Good point, thanks

Aberdeener:
Fixed

Aberdeener:
Updated

lucko:
Thanks!

Issue #3239, opened 2021-12-16
No error messages when exporting without filename


omit01:
### Description

When exporting permissions with the command `/lp export --without-users` and fogetting the name of the file to export to, there is no output to the user in the chat. But there is an error in the console because there is no name given

### Reproduction Steps

1. Go into a server (in this case a Purpur 1.17. Hotfix 2) with luckperms
2. Try to export with `/lp export --without-users`, don't give a name for the file
3. Wonder why you don't get feedback in chat and look in the console to see the error

### Expected Behaviour

An error is given in chat where Luckperms says: Command Usage ....

### Server Details

purpur-1.17.1-hotfix2

### LuckPerms Version

V5.3.70

### Logs and Configs

https://gist.github.com/omit01/39ecc407d77d971f488b29be3b82bf75

lucko:
This is now resolved, thanks to @OmegaWeaponDev

PR #3248, opened 2021-12-23
Add default file name for export command

OmegaWeaponDev:
Adds a default file name to the export command for when one isn't provided. This will resolve issue #3239

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/commands/misc/ExportCommand.java
+++ b/common/src/main/java/me/lucko/luckperms/common/commands/misc/ExportCommand.java
@@ -47,7 +47,7 @@ public class ExportCommand extends SingleCommand {
     private final AtomicBoolean running = new AtomicBoolean(false);
 
     public ExportCommand() {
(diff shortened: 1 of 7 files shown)


common/src/main/java/me/lucko/luckperms/common/commands/misc/ExportCommand.java
```diff
@@ -68,7 +71,16 @@ public void execute(LuckPermsPlugin plugin, Sender sender, ArgumentList args, St
             exporter = new Exporter.WebUpload(plugin, sender, includeUsers, includeGroups, label);
         } else {
             Path dataDirectory = plugin.getBootstrap().getDataDirectory();
-            Path path = dataDirectory.resolve(args.get(0) + ".json.gz");
+            Path path;
```
Aberdeener, line 74: Initializing the path variable here to what’s on line 82 would remove the need for the else statement

lucko: Won't work, `args.get(0)` will throw index out of bounds

Issue #3238, opened 2021-12-15
Full support for transient permissions through Vault API


Sataniel98:
### Description

LuckPerms does not implement the methods of Vault's permission API related to transient permissions:

* public boolean playerAddTransient(OfflinePlayer player, String permission)
* public boolean playerAddTransient(Player player, String permission)
* public boolean playerAddTransient(String worldName, OfflinePlayer player, String permission)
* public boolean playerAddTransient(String worldName, Player player, String permission)
* public boolean playerRemoveTransient(String worldName, OfflinePlayer player, String permission)
* public boolean playerRemoveTransient(String worldName, Player player, String permission)

See:

* https://github.com/LuckPerms/LuckPerms/blob/master/bukkit/src/main/java/me/lucko/luckperms/bukkit/vault/LuckPermsVaultPermission.java
* https://github.com/LuckPerms/LuckPerms/blob/master/bukkit/src/main/java/me/lucko/luckperms/bukkit/vault/AbstractVaultPermission.java
* https://github.com/MilkBowl/VaultAPI/blob/master/src/main/java/net/milkbowl/vault/permission/Permission.java#L203

It seems when these methods are called with LuckPerms, the add methods work but the remove methods don't "thanks" to the default implementation.

### Reproduction Steps

The problem was brought to my attention by users of my plugin and I confirmed the issue isn't on my end with a permission plugin that overrides the methods.

### Expected Behaviour

Removal of transient permissions

### Server Details

3360-Spigot-f65de01-c2e0bbd

### LuckPerms Version

5.3-SNAPSHOT (source code master)

lucko:
> the add methods work but the remove methods don't "thanks" to the default implementation.

This is arguably something Vault should fix 😛 

I'll consider fixing it in LP though, just got to find some time. PRs accepted in the meantime if anyone wants to give it a try, should be an good first issue.

lucko:
This is now resolved, thanks to @CuddleTheif

PR #3258, opened 2021-12-31
Added full support for tansient permissions through Vault API

CuddleTheif:
As per this [issue](https://github.com/LuckPerms/LuckPerms/issues/3238) I have added the missing methods that the Vault API does wrong and implemented them to use the transient Node on the user's Permission Holder.


lucko:
Looks great! Thank you!

I made a couple of changes, removed the unnecessary casts to `OfflinePlayer` (not your fault, they were there before for some reason) and implemented `playerAddTransient(OfflinePlayer player, String permission)` and `playerRemoveTransient(OfflinePlayer player, String permission)`.

Thanks again 😃

PR #3262, opened 2022-01-02
Forge

LXGaming:
This PR adds support for the [MinecraftForge](https://github.com/MinecraftForge/MinecraftForge) platform, the purpose of this is to be able to run LuckPerms without the need to use alternative platforms such as Bukkit / Forge Hybrids or Sponge.


common/src/main/java/me/lucko/luckperms/common/locale/TranslationManager.java
```diff
@@ -132,6 +132,15 @@ private void loadFromResourceBundle() {
         }
     }
 
+    private ResourceBundle getResourceBundle() {
+        try {
+            return ResourceBundle.getBundle("luckperms", DEFAULT_LOCALE, UTF8ResourceBundleControl.get());
+        } catch (UnsupportedOperationException ex) {
+            // ResourceBundle.Control not supported in named modules
```
lucko, line 139: What's the deal here? At the very least could we add a comment referencing the problem so regressions don't happen in future.

The UTF8ResourceBundleControl is necessary, as suggested, to make sure that the file is loaded with the correct encoding. (this wasn't the case on early Java versions)

We could switch the whole method to load the messages file as an InputStream/Properties object instead if that's more compatible.

forge/src/main/java/me/lucko/luckperms/forge/ForgeClassPathAppender.java
```diff
@@ -0,0 +1,135 @@
...
+import java.nio.file.Path;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+
+public class ForgeClassPathAppender extends URLClassLoader implements ClassPathAppender {
```
lucko, line 49: I'm not really sure what's going on here.

Ordinarily, `ClassPathAppenders` are not themselves class loaders - they just facilitate adding jars into the plugin/mod classpath - would be good to add some javadoc to clarify this :)

forge/src/main/java/me/lucko/luckperms/forge/ForgeClassPathAppender.java
```diff
@@ -0,0 +1,135 @@
...
+
+    @Override
+    protected Class<?> findClass(String name) throws ClassNotFoundException {
+        if (invalidClasses.contains(name)) {
+            throw new ClassNotFoundException(name);
+        }
+
+        invalidClasses.add(name);
+        Class<?> validClass = super.findClass(name);
+        invalidClasses.remove(name);
```
lucko, line 69: The `add()` followed by `remove()` seems strange - would a try+catch with a call to `add()` in the catch block not work? I think that would be more "correct"

forge/src/main/java/me/lucko/luckperms/forge/ForgeClassPathAppender.java
```diff
@@ -0,0 +1,135 @@
...
+        return validClass;
+    }
+
+    @Override
+    public void addJarToClasspath(Path file) {
```
lucko, line 74: Some comments to explain what's going on in this method would be good!

forge/src/main/java/me/lucko/luckperms/forge/ForgeClassPathAppender.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        if (this.parentLoaders == null) {
+            return;
+        }
+
+        try (JarFile jarFile = new JarFile(file.toFile(), false)) {
+            for (Enumeration<JarEntry> enumeration = jarFile.entries(); enumeration.hasMoreElements(); ) {
+                JarEntry jarEntry = enumeration.nextElement();
+                if (jarEntry.isDirectory() || !jarEntry.getName().endsWith(".class")) {
```
lucko, line 91: I don't think this approach will work with native libraries or other required resources?

I'm pretty sure some of the JDBC drivers need that

forge/src/main/java/me/lucko/luckperms/forge/ForgeClassPathAppender.java
```diff
@@ -0,0 +1,135 @@
...
+    @Override
+    public void close() {
+        // no-op
+    }
+
+    @SuppressWarnings("unchecked")
+    private Map<String, ClassLoader> getParentLoaders(TransformingClassLoader classLoader) {
+        try {
+            // sun.misc.Unsafe.theUnsafe
+            Field unsafeField = Unsafe.class.getDeclaredField("theUnsafe");
```
lucko, line 120: I would strongly prefer to not need to use `Unsafe` for this

forge/src/main/java/me/lucko/luckperms/forge/ForgeCommandExecutor.java
```diff
@@ -0,0 +1,111 @@
...
+import me.lucko.luckperms.common.sender.Sender;
+import net.minecraft.commands.CommandSourceStack;
+import net.minecraft.commands.Commands;
+import net.minecraftforge.event.RegisterCommandsEvent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+public class ForgeCommandExecutor extends CommandManager implements Command<CommandSourceStack>, SuggestionProvider<CommandSourceStack> {
```
lucko, line 47: Could we potentially move some of the generic brigadier stuff to the common module so Fabric/Forge can share it?

forge/src/main/java/me/lucko/luckperms/forge/ForgeSchedulerAdapter.java
```diff
@@ -0,0 +1,63 @@
...
+        public ForkJoinWorkerThread newThread(ForkJoinPool pool) {
+            return new WorkerThread(pool);
+        }
+    }
+
+    private static final class WorkerThread extends ForkJoinWorkerThread {
+
+        private WorkerThread(ForkJoinPool pool) {
+            super(pool);
+            setContextClassLoader(Thread.currentThread().getContextClassLoader());
```
lucko, line 59: Why is this necessary?

Some comments would be good here.

forge/src/main/java/me/lucko/luckperms/forge/ForgeSenderFactory.java
```diff
@@ -0,0 +1,110 @@
...
+            locale = null;
+        }
+
+        sender.sendSuccess(toNativeText(TranslationManager.render(message, locale)), false);
+    }
+
+    @Override
+    protected Tristate getPermissionValue(CommandSourceStack commandSource, String node) {
+        if (commandSource.getEntity() instanceof ServerPlayer) {
+            return ((ServerPlayerBridge) commandSource.getEntity()).bridge$hasPermission(node);
```
lucko, line 82: Can we use the Forge permissions API here instead?

Would be good to minimize Mixin use, as they are the main thing that cause problems on game/Forge version updates.

forge/src/main/java/me/lucko/luckperms/forge/ForgeSenderFactory.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Override
+    protected void performCommand(CommandSourceStack sender, String command) {
+        sender.getServer().getCommands().performCommand(sender, command);
+    }
+
+    @Override
+    protected boolean isConsole(CommandSourceStack sender) {
+        return sender.getEntity() == null;
```
lucko, line 103: Are you sure this is correct?

`commandSource.getEntity() != null && name.equals("Server")` is used above to check for console? Although I'm not convinced that's the correct way either..

forge/src/main/java/me/lucko/luckperms/forge/ForgeSenderFactory.java
```diff
@@ -0,0 +1,110 @@
...
+        if (commandSource.getEntity() != null) {
+            return commandSource.getEntity().getUUID();
+        }
+        return Sender.CONSOLE_UUID;
+    }
+
+    @Override
+    protected String getName(CommandSourceStack commandSource) {
+        String name = commandSource.getTextName();
+        if (commandSource.getEntity() != null && name.equals("Server")) {
```
lucko, line 61: What about a player with the username "Server" or an entity with that name.

Could we do an instanceof check instead?

forge/src/main/java/me/lucko/luckperms/forge/LPForgeBootstrap.java
```diff
@@ -0,0 +1,265 @@
...
+    public SchedulerAdapter getScheduler() {
+        return this.schedulerAdapter;
+    }
+
+    @Override
+    public ClassPathAppender getClassPathAppender() {
+        return this.classPathAppender;
+    }
+
+    // lifecycle
```
lucko, line 131: Do the lifecycle listeners support the integrated server correctly?

I think the fabric module is explicit about only enabling for the dedicated server - that's fine with me, but I think the code should enforce it to prevent any confusion.

forge/src/main/java/me/lucko/luckperms/forge/event/ConnectionEvent.java
```diff
@@ -0,0 +1,64 @@
...
+ *  SOFTWARE.
+ */
+
+package me.lucko.luckperms.forge.event;
+
+import com.mojang.authlib.GameProfile;
+import net.minecraft.network.Connection;
+import net.minecraftforge.eventbus.api.Event;
+
+public abstract class ConnectionEvent extends Event {
```
lucko, line 32: Aaaaah a shame that Forge doesn't have these events built-in.. that's a shame.

Do you know if there's any open feature requests for the functionality?

forge/src/main/java/me/lucko/luckperms/forge/messaging/PluginMessageMessenger.java
```diff
@@ -0,0 +1,103 @@
...
+        this.channel.addListener(event -> {
+            String message = event.getPayload().readUtf();
+            this.consumer.consumeIncomingMessageAsString(message);
+            event.getSource().get().setPacketHandled(true);
+        });
+    }
+
+    @Override
+    public void close() {
+        // no-op
```
lucko, line 73: I think the listener can probably be removed, at least??

forge/src/main/java/me/lucko/luckperms/forge/mixin/core/commands/CommandsMixin.java
```diff
@@ -0,0 +1,154 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.Redirect;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+
+@Mixin(value = Commands.class)
+public abstract class CommandsMixin {
```
lucko, line 52: Some javadocs/comments in this class would be useful.

Same goes for all of the mixins really 😄

forge/src/main/java/me/lucko/luckperms/forge/service/ForgePermissionHandler.java
```diff
@@ -0,0 +1,78 @@
...
+import net.minecraftforge.server.permission.nodes.PermissionNode;
+import net.minecraftforge.server.permission.nodes.PermissionTypes;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.UUID;
+
+public class ForgePermissionHandler implements IPermissionHandler {
```
lucko, line 44: Can we support contexts here?

Additionally, we should be able to support "meta" (prefix, suffix etc) lookups for the other permissiontypes.

Finally, can probably support offline permission lookup if the user data is cached.

lucko (commented):
Hello! 👋 

Many many thanks for the pull request. You've done a great job!

I've left some comments below. Lots of things to consider with new platforms, so please accept my apologies for so many questions!

Main sticking points for me are:
* **Dependency loading** - current impl would be fine if we can guarentee it will never break, but the current Java trends are tending towards the removal of hacks like Unsafe. I think with a bit of tweaking we can find a better way?
* **Permission API support** - I know it's a relatively new addition to Forge, but I think it's important to get the integration right first time. I have some reservations about the design choices but I hope that those can be worked through.
* **Lots of mixins** - well quite a few anyway! I'm keen to minimise them as much as possible, even if it means PRing new APIs to Forge.. that would be a much much better solution.

I'm happy to help resolve the above points! but of course interested to hear your initial replies/comments too.

Looking forward to seeing this finished & merged in 🎉

common/src/main/java/me/lucko/luckperms/common/locale/TranslationManager.java
LXGaming: I believe that [modlauncher](https://github.com/McModLauncher/modlauncher) loads everything as modules which leads to `ResourceBundle.Control` throwing an `UnsupportedOperationException` as it doesn't support being called from a named module, I'm unsure as to why the `ResourceBundle.Control` has this limitation, however it should only affect Java 9+

https://docs.oracle.com/javase/9/docs/api/java/util/ResourceBundle.Control.html#note
"ResourceBundle.Control is not supported in named modules. If the ResourceBundle.getBundle method with a ResourceBundle.Control is called in a named module, the method will throw an UnsupportedOperationException."

> We could switch the whole method to load the messages file as an InputStream/Properties object instead if that's more compatible.

This may be the best way forwards as other platforms may switch to using the Java Module System and may run into the same issue.

forge/src/main/java/me/lucko/luckperms/forge/ForgeClassPathAppender.java
LXGaming: During development `findClass` could end up being recursive which would lead to a StackOverflowError. I re-tested and this doesn't seem to be the case anymore, this is just left overs from when I was testing different Classloaders and could be changed.

forge/src/main/java/me/lucko/luckperms/forge/ForgeClassPathAppender.java
LXGaming: It was easier to just combine them due to hacks needed for [modlauncher](https://github.com/McModLauncher/modlauncher), I can separate them if need be.

forge/src/main/java/me/lucko/luckperms/forge/ForgeClassPathAppender.java
LXGaming: Unfortunately there isn't another option unless modlauncher is open to allowing mods to dynamically add jars at runtime, I did look into using Forges IModLocator but that requires the implementation to know ALL of the jars that need to be loaded in advanced, From what I can tell LuckPerms doesn't know this until after the configuration has been read but that requires loading configuration jars first, One potential solution is to ship the Forge build of LuckPerms with the dependencies required to read the configuration so that the rest of the required jars are known and can be loaded.

This is what happens with normal reflection:
`java.lang.reflect.InaccessibleObjectException: Unable to make field private final java.util.Map cpw.mods.cl.ModuleClassLoader.parentLoaders accessible: module cpw.mods.securejarhandler does not "opens cpw.mods.cl" to module luckperms`

forge/src/main/java/me/lucko/luckperms/forge/ForgeClassPathAppender.java
LXGaming: The parentLoaders is used by modlauncher for loading classes not resources.

Resources including JDBC drivers are loaded fine (Tested H2 & MySQL) as they call `getResourceAsStream` which uses the ClassLoader which loaded the class, in the case of H2 it's the `IsolatedClassLoader` and for MySQL it's the `ForgeClassPathAppender` as these ClassLoaders extend URLClassLoader they themselves handling pulling the resources from the jars.

forge/src/main/java/me/lucko/luckperms/forge/ForgeClassPathAppender.java
LXGaming: Comments added

forge/src/main/java/me/lucko/luckperms/forge/ForgeSchedulerAdapter.java
LXGaming: Comments added, I originally documented the reason for this in the commit message https://github.com/LuckPerms/LuckPerms/pull/3262/commits/ec8b8565977a22d42883278968f1f84bed3e93a9

forge/src/main/java/me/lucko/luckperms/forge/LPForgeBootstrap.java
LXGaming: Yes, I've tested running LuckPerms on the integrated server and it worked fine, I can make it server side only if you'd prefer.

forge/src/main/java/me/lucko/luckperms/forge/ForgeSenderFactory.java
LXGaming: The Forge [PermissionAPI.getPermission](https://github.com/MinecraftForge/MinecraftForge/blob/335cb5edc4467a213a01a315e75aad5aa3569f34/src/main/java/net/minecraftforge/server/permission/PermissionAPI.java#L67-L81) requires that the node being looked up has been registered. Not sure if this is doable

UeberallGebannt:
Are there any news about how this is going?

Razernok:
Are there plans to make this actually happen?

UeberallGebannt:
Would be very nice since Velocity is currently debating adding forge support for modern FML

Razernok:
sadly, hybrid servers like mohist, magma, or spongeforge hate modpacks. i always had issues like players unable to login.
i tried going that route first so i could have a few plugins like luckperms.

UeberallGebannt:
Well personally, I'm running ~5 modpack servers on Magma 1.12, and had much more running without issues.
However as hybrids are nowhere near playable on 1.16+ for larger servers, having a native Forge mod would be much better

lucko:
@LXGaming Thank you for resolving lots of my comments and merging master back in!

I will try to take another look soon, I think we're pretty close! :D

forge/src/main/java/me/lucko/luckperms/forge/mixin/core/commands/CommandsMixin.java
```diff
@@ -0,0 +1,91 @@
...
+        map.put(event.getNode(), suggestionNode);
+
+        fillUsableCommands(event.getNode(), suggestionNode, source.withPermission(4), map);
+    }
+
+    /**
+     * Forge will only re-throw the Exception passed in the {@link net.minecraftforge.event.CommandEvent CommandEvent} if it's an unchecked Exception,
+     * this behaviour prevents a {@link CommandSyntaxException CommandSyntaxException} from being propagated through the event.
+     *
+     * TODO Fix upstream.
```
LXGaming, line 76: This Mixin can be removed when this is merged: https://github.com/MinecraftForge/MinecraftForge/pull/8590

forge/src/main/java/me/lucko/luckperms/forge/event/ConnectionEvent.java
LXGaming: It appears that a PR for something very similar was closed due to the use cases (including LuckPerms use case of preloading data) needing more justification.

I'm assuming that due to this Forge is not open to this functionality.

https://github.com/MinecraftForge/MinecraftForge/pull/8393

forge/src/main/java/me/lucko/luckperms/forge/LPForgeBootstrap.java
lucko: Nope, this is fine :)

forge/src/main/java/me/lucko/luckperms/forge/ForgeClassPathAppender.java
lucko: @LXGaming I think it might make it clearer if they were separate?

forge/src/main/java/me/lucko/luckperms/forge/mixin/core/server/network/ServerLoginPacketListenerImplMixin.java
```diff
@@ -0,0 +1,100 @@
...
+        if (!this.gameProfile.isComplete()) {
+            this.gameProfile = this.createFakeProfile(this.gameProfile);
+        }
+
+        if (this.luckperms$authenticated) {
+            return;
+        }
+
+        this.luckperms$authenticated = true;
+        MinecraftForge.EVENT_BUS.post(new ConnectionEvent.Auth(this.connection, this.gameProfile));
```
lucko, line 83: Just to confirm - we can safely perform blocking calls here without disrupting the server thread?

LXGaming: This would run on the server thread however the only usage for this event is to allow LuckPerms to setup a `CompletableFuture` for asynchronously preloading player data while the Forge negotiation takes place, there is no blocking operation.

On `PlayerEvent.LoadFromFile` the `CompletableFuture` is blocked until completion, This might actually be an issue although the solution would to replace `get()` with `getNow(false)` so that throws if it's not completed.

I could remove these events entirely however there would need to be a statically accessible instance of LPForgePlugin as it's not possible to obtain it.

https://github.com/LuckPerms/LuckPerms/blob/ab07f1f5391534a199ba38e195f6df9c7263e04f/forge/src/main/java/me/lucko/luckperms/forge/listeners/ForgeConnectionListener.java#L59-L94
https://github.com/LuckPerms/LuckPerms/blob/ab07f1f5391534a199ba38e195f6df9c7263e04f/forge/src/main/java/me/lucko/luckperms/forge/listeners/ForgeConnectionListener.java#L106-L111

forge/src/main/java/me/lucko/luckperms/forge/listeners/ForgeConnectionListener.java
```diff
@@ -0,0 +1,161 @@
...
+                this.plugin.getEventDispatcher().dispatchPlayerLoginProcess(profile.getId(), profile.getName(), null);
+                return !this.plugin.getConfiguration().get(ConfigKeys.CANCEL_FAILED_LOGINS);
+            }
+        }, this.plugin.getBootstrap().getScheduler().async());
+
+        this.pendingConnections.put(profile.getId(), future);
+    }
+
+    @SubscribeEvent
+    public void onPlayerLoadFromFile(PlayerEvent.LoadFromFile event) {
```
LXGaming, line 97: TODO Fix: This event gets fired BEFORE the `player.connection` field has been set and results in the catch logic failing.
This may unfortunately require another Mixin as Forge doesn't have any events that fit.

forge/src/main/java/me/lucko/luckperms/forge/ForgeClassPathAppender.java
LXGaming: I've figured out how to avoid using `Unsafe` by reviewing how Bukkit, Bungee and Sponge are loaded via the `LoaderBootstrap`.

The only downside I've found so far is that Forge events have to be manually registered due to the ASM that Forge generates not taking the event classes classloader in account which results in ClassNotFoundExceptions on event invocation.

[LXGaming/LuckPerms@forge-loader](https://github.com/LXGaming/LuckPerms/commit/263acdac710909dfcc3c50ed3396ee7ddc1a2fa2)

forge/src/main/java/me/lucko/luckperms/forge/mixin/core/server/network/ServerLoginPacketListenerImplMixin.java
lucko: Hmm `getNow(false)` isn't going to work.

We need to block the player connection handling while the database load is performed. This could take >100ms, so it's really important that we block and wait, and not just "hope" that it completes async by the time we need it.

All other platforms have an event or some sort of construct for this: 

https://github.com/LuckPerms/LuckPerms/blob/c35958b88ec05022270461d2b8c522c8d991a008/bukkit/src/main/java/me/lucko/luckperms/bukkit/listeners/BukkitConnectionListener.java#L91

https://github.com/LuckPerms/LuckPerms/blob/c35958b88ec05022270461d2b8c522c8d991a008/fabric/src/main/java/me/lucko/luckperms/fabric/listeners/FabricConnectionListener.java#L80-L81

You can see the Fabric impl with Mixins here: https://github.com/FabricMC/fabric/tree/1.18.2/fabric-networking-api-v1

https://github.com/FabricMC/fabric/blob/1.18.2/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/mixin/networking/ServerLoginNetworkHandlerMixin.java#L59-L65
https://github.com/FabricMC/fabric/blob/1.18.2/fabric-networking-api-v1/src/main/java/net/fabricmc/fabric/impl/networking/server/ServerLoginNetworkAddon.java#L70-L111

My strong preference would be to request + wait for this functionality to exist in Forge. I'm relucant to hack our own hooks in with Mixins for this sort of thing...

LXGaming: Hey @lucko, I was wondering if you could provide your thoughts and reasoning as to why LuckPerms needs this event added to Forge and why consumers of LuckPerms need the user data early.
https://github.com/MinecraftForge/MinecraftForge/pull/8599

joaoespanhol:
forge approved pull request  https://github.com/MinecraftForge/MinecraftForge/pull/8599
how have feedback?

joaoespanhol:
how have link for download?
for testing mod

LXGaming:
> how have link for download? for testing mod

There isn't one, It's still being actively developed.

lucko:
> how have link for download? for testing mod

@joaoespanhol Please stop pestering - it'll be ready when it's ready. Asking for updates isn't going to make things progress any faster.

common/src/main/java/me/lucko/luckperms/common/locale/TranslationManager.java
LXGaming: With the jar-in-jar system this is no longer an issue as the `TranslationManager` is no longer being loaded from a named module.

However it should be noted that this issue could reappear in the future as platforms advance to Java 9+ and adopt the Java Module System.

forge/src/main/java/me/lucko/luckperms/forge/ForgeCommandExecutor.java
LXGaming: Good idea, Didn't realize Fabric added command selector support for LuckPerms commands via this system.
Any thoughts on this https://github.com/LuckPerms/LuckPerms/pull/3262/commits/2f5b44e16763bc87c4aaac6cc5e5161310b0505a I've common the Fabric and Forge brigadier logic together as best as I could, The `resolveSelectors` has to remain platform dependent due to the different Minecraft mappings.

I created the `BrigadierCommandExecutor` as I didn't want to implement brigadier directly into the `CommandManager` incase it broke platforms that don't have or use brigadier.

lucko: Yes, looks good!

common/src/main/java/me/lucko/luckperms/common/http/AbstractHttpClient.java
```diff
@@ -51,4 +52,15 @@ protected Response makeHttpRequest(Request request) throws IOException, Unsucces
         }
         return response;
     }
+
+    public void shutdown() {
```
lucko, line 56: This change was implemented a little differently on the master branch recently -- 5647b2e5c6b13f685f492e49e4af83d7ffa402fa

If you rebase I think this can be removed :)

common/src/main/java/me/lucko/luckperms/common/plugin/scheduler/WorkerThreadFactory.java
```diff
@@ -0,0 +1,49 @@
...
+        return new WorkerThread(pool);
+    }
+
+    private static final class WorkerThread extends ForkJoinWorkerThread {
+
+        private WorkerThread(ForkJoinPool pool) {
+            super(pool);
+            // By default, the ContextClassLoader is set to the SystemClassLoader, This causes ClassNotFoundExceptions
+            // due to modlauncher loading LuckPerms on its TransformingClassLoader,
+            // we must replace the ContextClassLoader with the current one to ensure the correct ClassLoader is used.
```
lucko, line 44: Is this still needed now we're using jar-in-jar?

forge/src/main/java/me/lucko/luckperms/forge/messaging/PluginMessageMessenger.java
```diff
@@ -0,0 +1,98 @@
...
+            ServerPlayer player = this.plugin.getBootstrap().getServer()
+                    .map(MinecraftServer::getPlayerList)
+                    .map(PlayerList::getPlayers)
+                    .map(players -> Iterables.getFirst(players, null))
+                    .orElse(null);
+            if (player == null) {
+                return;
+            }
+
+            FriendlyByteBuf byteBuf = new FriendlyByteBuf(Unpooled.buffer());
```
lucko, line 84: Recently discovered that the utf8 format used by FriendlyByteBuf isn't quite the same as what LP expects.

If we rebase, can switch to use the newly added `AbstractPluginMessageMessenger` - copy the change made to Fabric: https://github.com/LuckPerms/LuckPerms/commit/c35958b88ec05022270461d2b8c522c8d991a008#diff-022aacd2913f3d175268a7ede8f4176df3b1e0739c96885d3599116a7595a2c7

forge/src/main/java/me/lucko/luckperms/forge/model/ForgeUser.java
```diff
@@ -0,0 +1,128 @@
...
+import me.lucko.luckperms.common.model.User;
+import me.lucko.luckperms.common.verbose.event.CheckOrigin;
+import net.luckperms.api.query.QueryOptions;
+import net.luckperms.api.util.Tristate;
+import net.minecraft.server.level.ServerPlayer;
+
+import java.util.Locale;
+import java.util.function.Function;
+
+public class ForgeUser {
```
lucko, line 41: Sorry.. I know we removed Mixins, but this seems like a good candidate... what do you think?

Failing that, the other option would be to use the InlineQueryOptionsSupplier, like BungeeCord does: https://github.com/LuckPerms/LuckPerms/blob/master/bungee/src/main/java/me/lucko/luckperms/bungee/context/BungeeContextManager.java#L64

I'd prefer that as opposed to having an additional map of users in the context manager

forge/src/main/java/me/lucko/luckperms/forge/util/EventBusUtil.java
```diff
@@ -0,0 +1,95 @@
...
+import java.lang.invoke.CallSite;
+import java.lang.invoke.LambdaMetafactory;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.function.Consumer;
+
+public class EventBusUtil {
```
lucko, line 41: Please could you add some JavaDocs here to explain what's happening / why it is necessary? :)

lucko (changes requested):
Getting very close! Exciting stuff! 😄 

Just a few more comments / suggestions for your consideration. :)

forge/src/main/java/me/lucko/luckperms/forge/model/ForgeUser.java
LXGaming: I've switched to using Forge Capabilities https://github.com/LuckPerms/LuckPerms/pull/3262/commits/21f3805cf3355cb346cf2e87dbb4af6d645f8597 the `UserCapability` (previously `ForgeUser`) is stored on the `ServerPlayer` object without the need for Mixins.

forge/src/main/java/me/lucko/luckperms/forge/util/EventBusUtil.java
LXGaming: Yeah absolutely, I'll be doing a pass over the changes to add documentation.

forge/src/main/java/me/lucko/luckperms/forge/service/ForgePermissionHandler.java
LXGaming: I've added context and offline permission / meta support, not entirely sure if this is the correct way to add contexts to a existing `QueryOptions`
https://github.com/LuckPerms/LuckPerms/blob/fed94e36fd55a919dff75a0a4d0b3e063f1d480d/forge/src/main/java/me/lucko/luckperms/forge/service/ForgePermissionHandler.java#L125-L134

Verbose:
- Online: https://luckperms.net/verbose/wyDwvVie8t
- Offline: https://luckperms.net/verbose/l80gEzDrxg

forge/src/main/java/me/lucko/luckperms/forge/model/ForgeUser.java
lucko: Perfect :+1:

lucko (approved):

Issue #3294, opened 2022-01-25
Support env var


HookWoods:
### Description

Hello,
It would be really useful to have support for variable env. Actually, my company uses Kubernetes to auto-deploy server and we actually use var env to configure database credentials because they are dynamically created each time we redeploy our database and we can't modify them in the config every time.

### Proposed Behaviour

At the start of the plugin, it could use the java env API to take the value from the var env, or the one from the config if not found.
`System.getenv().getOrDefault("LP_XX", ...);`
The only problem with the actual system would be the object cast.

lucko:
Closing in favour of your pull request

PR #3295, opened 2022-01-25
Add env variable option to read config

HookWoods:
This pull request adds an option to allow the user to use System env / variable env with the config.
Like that people that use dynamic servers like with Kubernetes can instead of writing data to config use variable env to read it.

I actually put it on String, Integer, and Boolean since it's useless with map and list.
Also, people need to add -Dluckperms.read-env=true, if this is false, it will use the default behavior of config, instead, it will check if the variable is defined in the env, or will read the config if not.

Associated with [3294](https://github.com/LuckPerms/LuckPerms/issues/3294)

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/config/generic/adapter/ConfigurateConfigAdapter.java
+++ b/common/src/main/java/me/lucko/luckperms/common/config/generic/adapter/ConfigurateConfigAdapter.java
@@ -43,6 +43,7 @@ public abstract class ConfigurateConfigAdapter implements ConfigurationAdapter {
     private final LuckPermsPlugin plugin;
     private final Path path;
     private ConfigurationNode root;
+    private static final boolean SHOULD_READ_ENV = Boolean.getBoolean("luckperms.read-env");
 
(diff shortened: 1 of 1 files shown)


lucko:
Hi, thanks for this.

I think the change is valid - but would prefer to see it implemented in a generic way. Currently, it is only implemented by the Configurate config adapter, so will only function on Sponge, Fabric and Velocity.

Once way to do it "properly" would be to create an abstract impl of `ConfigurationAdapter` with the env var code in it, then have the others extend it.

I'm also not sure about using environment variables vs system properties - I think I would prefer the latter, but I suppose both would also work.

clrxbl:
I'm in the same boat where I utilize Kubernetes to orchestrate servers at scale.
I currently have to inject secrets into the LuckPerms YAML config but I would like to just be able to configure this with an environment variable to make life easier.

Do you plan on continuing with this PR?

lucko:
Added in 126631fce8b60b5548ae5d932722e8bb667d28b1

https://luckperms.net/wiki/Configuration#system-properties
https://luckperms.net/wiki/Configuration#environment-variables

Thanks @HookWoods for the original PR :)

PR #3303, opened 2022-02-04
Web editor socket connection

lucko:
Please see below for a short demo. 😎 

https://user-images.githubusercontent.com/8352868/152601802-dea49f39-2121-4dce-b33e-907bd66d8824.mp4

## Overview

This change adds support for bidirectional communication between the web editor and the LuckPerms plugin, allowing edits made in the editor to be applied in one click. It is implemented using the [WebSocket protocol](https://datatracker.ietf.org/doc/html/rfc6455).

* The transfer of data between the LuckPerms plugin (runs on the users Minecraft server) and the LuckPerms editor (runs in the users web browser) is facilitated by a separate web service called [bytebin](https://github.com/lucko/bytebin).
  * This is the service that has always powered the web editor, as well as some of my other projects (spark, pastes.dev, etc).
  * The plugin posts permissions data to bytebin, which stores the data and returns a unique code to reference it.
  * The user is then prompted to open the editor with the unique code in the URL, which allows the editor to retrieve the posted data. A similar process occurs in reverse when the editor changes are saved. The user has to run a command containing a unique code on their server.
* A **new** service has been introduced called [bytesocks](https://github.com/lucko/bytesocks) (I'm bad at names ok) which is similar to bytebin, except that instead of using GET and POST requests, it allows clients to connect via a WebSocket to a bytesocks "channel" and send messages to other connected clients.
  * This allows for continuous bi-directional communication without extra action from the user.
  * However, this opens up some weak points in terms of security which must be considered carefully to ensure the system remains secure.

## Security
The following processes and checks have been implemented, which to the best of my knowledge, will ensure that the system is secure (or is at least as secure as the previous system).

#### 1) Signature verification using public/private keypairs

* Both the plugin and editor generate and store an RSA keypair
* The plugin public key is sent in the initial data payload, and the editor will only acccept messages from the socket if they have a valid signature created with the plugin key.
* The editor public key is sent when it first attempts a "handshake" with the server via the socket. The plugin maintains a store of trusted editor public keys. The user will need to manually confirm new browsers when they connect for the first time. The plugin will only accept messages from the socket if they have a valid signature created with a trusted editor key.

#### 2) Other safeguards

* The user who initiated the session must remain online on the server for changes to be applied automatically.
* The users permission status is re-checked each time an automatic change is received.

#### Reviews
If you have an understanding of app security, I'd really appreciate your help in this area! The main cryptographic code can be found in [`ws.js`](https://github.com/LuckPerms/LuckPermsWeb/blob/master/src/util/ws.js) in LuckPermsWeb and the [`socket`](https://github.com/LuckPerms/LuckPerms/tree/feature/editor-socket/common/src/main/java/me/lucko/luckperms/common/webeditor/socket) package in LuckPerms.

Please have a look at let me know if you think anything has been overlooked. :)

## Beta Testing
Most importantly, we need your help testing!

This is a big change and we'd like to catch any remaining bugs or problems before we merge it into the main release. Please download using the links below, unzip and test it out on your server!

* [LuckPerms-Bukkit-5.3.108.jar.zip](https://github.com/LuckPerms/LuckPerms/files/8006203/LuckPerms-Bukkit-5.3.108.jar.zip)
* [LuckPerms-Sponge-5.3.108.jar.zip](https://github.com/LuckPerms/LuckPerms/files/8006209/LuckPerms-Sponge-5.3.108.jar.zip)


LoJoSho:
:o

the1sword:
This is incredible! Amazing work!

Skjolberg:
Amazing dude, thanks

PR #3309, opened 2022-02-09
Optimize parent resolution in info commands

Sxtanna:
Reworks the `GroupInfo` and `UserInfo` commands' method of resolving permanent and temporary parent nodes.

Uses Java 8's "Grouping By" collector to group the nodes into a map of results, rather than executing the entire operation twice with a different filter.

*(i would actually have loved to have this operation be generic and reusable outside of these commands, but I don't see where things like that are stored in the project, so duplicate code is fine for now ig)*


lucko:
Looks great, thank you!

> *(i would actually have loved to have this operation be generic and reusable outside of these commands, but I don't see where things like that are stored in the project, so duplicate code is fine for now ig)*

Sometimes repeated code is better than a complicated abstraction - I think this is one of those cases :)

PR #3321, opened 2022-02-19
Upgrade gradle 7.2 -> 7.4

vie10:
Gradle 7.4 release notes: https://docs.gradle.org/7.4/release-notes.html


vie10:
We still can keep the gradle version up to date without any issues.
@lucko review it, as possible.

powercasgamer (changes requested):
You should also update the wrapper jar using `./gradlew wrapper --gradle-version 7.4`

vie10:
> You should also update the wrapper jar using `./gradlew wrapper --gradle-version 7.4`

Yep. Sorry for that, seems I've not included it in the first commit.

powercasgamer (approved):

lucko:
Unless this is actually fixing a problem with the buildscript, I'd prefer to not update. It creates a lot of unnecessary and unrelated noise in the LuckPerms commit log by constantly bumping the Gradle version. (frankly I'd prefer to not have a wrapper at all, but it seems to be the norm)

lucko:
Ok I'll take that as a no then :)

p.s I noticed your profile picture and location - wishing you all the best, stay safe.

vie10:
yep. sorry but I've not more time for do more right now (
thanks!

PR #3330, opened 2022-03-02
Fix Travertine download link

Aberdeener:
Travertine has now been completely discontinued, and only exists on the legacy downloads page. This PR updates the link to point to the correct page on the Paper website


lucko:
Hmm if it's been discontinued then I'm not sure we should be recommending it.

Issue #3339, opened 2022-03-16
Setup server name through the API and/or environment variable


schneidermr:
### Description

That would be great if there will be an option to setup the server name using the Luckperms API and/or using an environment variable.

### Proposed Behaviour

Need to create a setServerName function to support dynamic server name changing.
Oh and it would be also very cool if you could implement the possibility to setup the server name via environment variable and not only using runtime variable (it is nearly a must have option if users running the plugin in a Docker container).

lucko:
In your LuckPerms config, set:

```yml
server: load-from-system-property
```

then set the `luckperms.server` system property.

---

You can do this in Java code using: (must be called before LuckPerms loads)
```java
System.setProperty("luckperms.server", "specialserver");
```

or with a server startup argument:
```
java -Dluckperms.server=specialserver -jar server.jar
```

SlimeDog:
Please add commentary to the configuration file, explaining how to configure this feature. Does it replace `server: global` (for example)?

PR #3341, opened 2022-03-16
resolves LuckPerms/LuckPerms#3339

schneidermr:
With this change now anyone could be able setup the server name also via environment variable


lucko:
Thanks!

derklaro:
I think the assignment of the actual server variable is missing here... (just by looking at the code) 🤔

schneidermr:
> I think the assignment of the actual server variable is missing here... (just by looking at the code) 🤔

Oh holly... you're absolutely right, I'll fix it

PR #3360, opened 2022-04-08
Table prefix/collections display in '/lp info'

ZeromaniacMc:
Per my own request, my own PR;
I added the table prefix (for hikari)/collection (for mongodb) to the /lp info output.
The reason for it is quite simple: debugging. Unfortunately users sometimes still confuse tables and databases so this would be an easy way to debug data not syncing whilst multiple servers simply say "mysql" and "connected" but not where to.

P.S.: I really tried xD


common/src/main/java/me/lucko/luckperms/common/config/ConfigKeys.java
```diff
@@ -555,11 +555,16 @@ private ConfigKeys() {}
         );
     }));
 
+    /**
+     * Default value for table prefix
+     */
+    public static final String defaultTablePrefix = "luckperms_";
```
powercasgamer, line 561:
```suggestion
    public static final String DEFAULT_TABLE_PREFIX = "luckperms_";
```

ZeromaniacMc: Fair point. Changed. Thanks :)

powercasgamer (approved):

lucko:
Thanks!

PR #3363, opened 2022-04-21
Resolve command entity selectors on Fabric

emilyy-dev:


lucko:
Looks great, thank you!

PR #3403, opened 2022-06-24
Fix concurrent errors with unopenedURLs

OOP-778:
I had issues using ReflectionAppender with concurrency, so adding synchronized fixes it.


lucko:
Thanks, sorry it took a while to merge!

PR #3477, opened 2022-08-30
Forge 1.19.2

CDAGaming:
- One minor systems change to get ParseResults, but nothing that should be breaking, and resolves a NoSuchMethod error during server init.


lucko:
Thanks, sorry it took a while to merge!

PR #3504, opened 2022-10-10
add: Fabric offline permission check event

DrexHD:
Implements lucko/fabric-permissions-api#9

_(The changes in [lucko/fabric-permissions-api](https://github.com/lucko/fabric-permissions-api) have to be published before merging)_


lucko:
Thanks!

PR #3519, opened 2022-11-12
Minestom Support

HEROOSTECH:
It's time for it... Minestom LuckPerms. It's made by LeoDog896 (He allowed for me to create pr), i just updated it to latest minestom. This use minestom permission system and doesn't override hasPermission method, because minestom currently doesn't allow to do this (it's possible to add PlayerProvider, but most of servers will use custom PlayerProvider). Sorry for my bad English btw.

PR #3521, opened 2022-11-13
Minestom Support

Codestech1:


HEROOSTECH:
#3519 But with hydrogen's credit

underscore11code:
I think reposting [this](https://github.com/LuckPerms/LuckPerms/pull/3519#issuecomment-1312710888) comment from @dev-hydrogen is relevant
> That also ignores the issue that the way this works is at best untested, and at worst unstable. How i made my fork work, is that it (tries) to sync up permissions between luckperms and minestom when the player join, and then listens to the luckperms api to sync any changes that happen after that. This approach doesnt allow for any nice-to-haves like wildcard support, as that would mean needing to override hasPermission, which currently is not possible on minestom.
> 
> Theres a reason i never made a pull request from my fork, I don't trust it enough to be used anywhere other than on my own servers.

HEROOSTECH:
> I think reposting [this](https://github.com/LuckPerms/LuckPerms/pull/3519#issuecomment-1312710888) comment from @dev-hydrogen is relevant
> 
> > That also ignores the issue that the way this works is at best untested, and at worst unstable. How i made my fork work, is that it (tries) to sync up permissions between luckperms and minestom when the player join, and then listens to the luckperms api to sync any changes that happen after that. This approach doesnt allow for any nice-to-haves like wildcard support, as that would mean needing to override hasPermission, which currently is not possible on minestom.
> > Theres a reason i never made a pull request from my fork, I don't trust it enough to be used anywhere other than on my own servers.

Ik, but tell it for minestom, currently (if you merge) you can only note users about limitations. If minestom extensiona will add able to override hasPermission - i will do new pr. Lucko should choose: no minestom or minestom with limitations (e.g. no wildcards). Sorry for bad english

HEROOSTECH:
I have idea for wildcards, but i will do this in other pr (if you merge this).

Codestech1:
It's ready to use, you don't need wildcards, latest version of minestom has it built in.

lucko:
I'm still unsure whether this should be part of the original project, but I am definitely against merging it until Minestom sort out their Maven repository situation and/or start creating proper releases. 

The Jitpack releases expire after a while and are a source of annoyance in my other projects which support Minestom (e.g. spark)

xHyroM:
> It's ready to use, you don't need wildcards, latest version of minestom has it built in.

I tried this but when someone sets permission from another server it doesn't update permissions in minestom server so i can still use /lp commands

Bloeckchengrafik:
> It's ready to use, you don't need wildcards, latest version of minestom has it built in.

With those, negated permissions won't work (at least in my testing)

ajh123:
Well Minestom is now on Maven central, so we no longer need Jitpack. However there is no official extension support in the newest version so https://github.com/hollow-cube/minestom-ce-extensions has to be used.

MelonHell:
now Minestom does not use jitpack, only mavenCentral

ajh123:
> is there anything in the works to officialy add support for this platform with a library?

@Danieljunek17 I don't think there is any official work yet. I don't think the LuckPerms developers have interest at the moment (mainly because it is not as popular as compared to Spigot, however there is a new YouTube tutorial that might change that).

Codestech1:
I heard that Minestom is going to remove their permission API. Also, i'm not using Mienstom anymore and i don't plan on maintaing this pull request anymore.

ajh123:
@Codestech1 Where did you here that "Minestom is going to remove their permission API"? To me it looks like [permissions are going to be improved](https://github.com/Minestom/Minestom/issues/2021).

Codestech1:
> @Codestech1 Where did you here that "Minestom is going to remove their permission API"? To me it looks like [permissions are going to be improved](https://github.com/Minestom/Minestom/issues/2021).

On their discord iirc

TheMeinerLP:
Our fork will continue with extension and permission fork for everyone

PR #3525, opened 2022-11-20
Adding Nats messenger implementation to LuckPerms

sadcenter:
Adds [nats](https://nats.io/) implementation to LuckPerms messaging system.
also, removes message about loading and not recognizing messenger when it is set to `custom`.

### DIFF
--- a/bukkit-legacy/build.gradle
+++ b/bukkit-legacy/build.gradle
@@ -34,6 +34,7 @@ shadowJar {
     relocate 'com.mongodb', 'me.lucko.luckperms.lib.mongodb'
     relocate 'org.bson', 'me.lucko.luckperms.lib.bson'
     relocate 'redis.clients.jedis', 'me.lucko.luckperms.lib.jedis'
+    relocate 'io.nats.client', 'me.lucko.luckperms.lib.nats'
     relocate 'com.rabbitmq', 'me.lucko.luckperms.lib.rabbitmq'
     relocate 'org.apache.commons.pool2', 'me.lucko.luckperms.lib.commonspool2'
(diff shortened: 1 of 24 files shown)


lucko:
Looks good - happy to accept this. :)

Have you tested it on a server with just LuckPerms installed?

Seems like the dependency is missing. You'll need a new entry in the [Dependency class](https://github.com/LuckPerms/LuckPerms/blob/master/common/src/main/java/me/lucko/luckperms/common/dependencies/Dependency.java) and also probably a new condition [here in the DependencyRegistry](https://github.com/LuckPerms/LuckPerms/blob/master/common/src/main/java/me/lucko/luckperms/common/dependencies/DependencyRegistry.java#L91) class.

sadcenter:
yeah I forgot to add dependency, I thought that everything is ok because my plugin is shading nats.
Now everything should be working as intended. Here is a screenshot from velocity server with nats enabled.
image

lucko:
Thanks :)

I've also done some quick testing - all looks good!

```yml
version: '3'

services:
  luckperms-a:
    image: luckperms
    environment:
      LUCKPERMS_MESSAGING_SERVICE: nats
      LUCKPERMS_NATS_ENABLED: true
      LUCKPERMS_NATS_ADDRESS: 'nats'

  luckperms-b:
    image: luckperms
    environment:
      LUCKPERMS_MESSAGING_SERVICE: nats
      LUCKPERMS_NATS_ENABLED: true
      LUCKPERMS_NATS_ADDRESS: 'nats'

  nats:
    image: nats
```

Screenshot 2022-11-26 at 18 02 26

sadcenter:
also don't forget about https://github.com/LuckPerms/wiki/pull/144

lucko:
Ah yep, thanks for the reminder!

PR #3541, opened 2022-12-21
Added documentation

Clouke:


lucko:
Hi, appreciate the PR but not sure this is really a necessary addition! (those methods are fairly self explainatory, and it's an internal (not API/user-facing) class.

PR #3565, opened 2023-01-19
Add nullcheck to UUIDs

underscore11code:
Adds a null check for UUIDs on join. Theoretically this shouldn't be possible, but some offline mode authentication plugins and server softwares with offline mode are...special. Adding an explicit check instead of waiting for it to eventually hit something that complains also lets us also have a friendlier error for the end user to read so they can (hopefully) self-diagnose,

I haven't tested the code besides ensuring it builds (partially because I'm not actually sure how to easily cause a null UUID), but I'd like to think I know how to write a null check without screwing up.

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/plugin/util/AbstractConnectionListener.java
+++ b/common/src/main/java/me/lucko/luckperms/common/plugin/util/AbstractConnectionListener.java
@@ -27,7 +27,6 @@ package me.lucko.luckperms.common.plugin.util;
 
 import me.lucko.luckperms.common.model.User;
 import me.lucko.luckperms.common.plugin.LuckPermsPlugin;
-
 import net.luckperms.api.model.PlayerSaveResult;
 import net.luckperms.api.model.data.DataType;
 import net.luckperms.api.platform.Platform;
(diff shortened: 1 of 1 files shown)


common/src/main/java/me/lucko/luckperms/common/plugin/util/AbstractConnectionListener.java
```diff
@@ -61,6 +60,11 @@ protected void recordConnection(UUID uniqueId) {
     }
 
     public User loadUser(UUID uniqueId, String username) {
+        if (uniqueId == null) {
+            this.plugin.getLogger().severe("Attempted to load data for player '" + username +"' with a null UUID.! This implies a plugin/mod/your server software is interfering with UUIDS.");
```
powercasgamer, line 64:
```suggestion
            this.plugin.getLogger().severe("Attempted to load data for player '" + username + "' with a null UUID! This implies a plugin/mod/your server software is interfering with UUIDS.");
```

powercasgamer (changes requested):

powercasgamer (approved):

LeotomasMC:
> because I'm not actually sure how to easily cause a null UUID
fabric in offline mode

Logics4:
> > because I'm not actually sure how to easily cause a null UUID
> 
> fabric in offline mode

I'm completely out of the loop with how Fabric works, so let me ask, what is LuckPerms doing of different that it somehow complains about a null UUID in offline mode? In Bukkit, Sponge, and BungeeCord, LuckPerms seems to just work with whatever the platform gives them (`getUniqueId()` will return a UUID, period. Whether it is v3 [offline mode] or v4 [online mode], it doesn't specify). Does the Fabric server assign a UUIDv3 too late in the login lifecycle for LuckPerms to do anything with it?

LeotomasMC:
luckperms does just use the platforms method to get the player's uuid, but for whatever reason, the uuid is null at that stage on fabric in offline mode. this is more likely a side effect of that being how vanilla behaves as fabric does not alter vanilla behavior on its own (thats what mods are for!). although i cant confirm if this happens without luckperms, the fact that the uuid is null in the log line showing the kick message indicates to me that its not because of luckperms.

just for the sake of it, [heres](https://pastes.dev/aO9nvOpnb7) the actual error that occurs

underscore11code:
Underlying issue solved by #3576, don't see a need for this anymore.

PR #3583, opened 2023-02-07
Update mariadb jdbc driver to 3.1.2

Kamillaova:
In older versions of the driver (before 3.1.0?) there is a bug, in newer versions I have not noticed it
(litebans uses 3.0.7)

image
image

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/storage/implementation/sql/connection/hikari/MariaDbConnectionFactory.java
+++ b/common/src/main/java/me/lucko/luckperms/common/storage/implementation/sql/connection/hikari/MariaDbConnectionFactory.java
@@ -49,23 +49,20 @@ public class MariaDbConnectionFactory extends HikariConnectionFactory {
 
     @Override
     protected void configureDatabase(HikariConfig config, String address, String port, String databaseName, String username, String password) {
-        config.setDataSourceClassName("org.mariadb.jdbc.MariaDbDataSource");
-        config.addDataSourceProperty("serverName", address);
-        config.addDataSourceProperty("port", port);
-        config.addDataSourceProperty("databaseName", databaseName);
+        config.setDriverClassName("org.mariadb.jdbc.Driver");
+        config.setJdbcUrl("jdbc:mariadb://" + address + ":" + port + "/" + databaseName);
         config.setUsername(username);
         config.setPassword(password);
     }
 
     @Override
-    protected void setProperties(HikariConfig config, Map<String, Object> properties) {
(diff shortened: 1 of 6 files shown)


common/src/main/java/me/lucko/luckperms/common/storage/implementation/sql/connection/hikari/MariaDbConnectionFactory.java
```diff
@@ -50,23 +50,10 @@ protected String defaultPort() {
 
     @Override
     protected void configureDatabase(HikariConfig config, String address, String port, String databaseName, String username, String password) {
-        config.setDataSourceClassName("org.mariadb.jdbc.MariaDbDataSource");
```
lucko, line 53: Is there a reason for switching to JDBC URL instead of the DataSource?

Kamillaova: > Is there a reason for switching to JDBC URL instead of the DataSource?

idk how to set databaseName with `MariaDbDataSource` in 3.x version

Kamillaova: @lucko

lucko: Should just be a case of finding the datasource class and defining it as we did previously.

It will implement the DataSource interface

Kamillaova: > Should just be a case of finding the datasource class and defining it as we did previously.
> 
> It will implement the DataSource interface

```java
config.addDataSourceProperty("databaseName", databaseName);
```

will throw error like "databaseName not found"

also, MariaDbDataSource from 2.x and MariaDbDataSource from 3.x are very different
image
and in 3.x there is no setProperties method, in 2.x it is deprecated


hmm, maybe just replace setDriverClassName with setDataSourceClassName?

Kamillaova: @lucko

lucko:
Hi @Kamillaova, sorry for the delay!

I have recently added some integration tests for the storage backends, please could you rebase your PR and ensure they pass? :) After that I think I am happy to merge this one

PR #3597, opened 2023-02-24
Add date on hover to log

### LINKED ISSUE #3408: Show full date in history
### Description

In the log command it would be good to see the whole date the action was taken on instead of just time since ago. 

### Proposed Behaviour

Currently we can only see how long ago was the action taken, which may not be enough if you need to see the exact date as you need to manually find the date by yourself. Especially when dealing with history thats older than just few days. So by adding for example a hover message on the currently displayed time since ago, with the full date, could be very usefull (especially when you cannot even get the history using the API to manage it yourself).

### Extra Details

_No response_

Diddyy:
This closes #3408.

On hover message on the currently displayed time since ago the full date and time is shown.
image

This is my first PR so please let me know if I have missed anything!

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/locale/Message.java
+++ b/common/src/main/java/me/lucko/luckperms/common/locale/Message.java
@@ -57,6 +57,8 @@ import net.luckperms.api.util.Tristate;
 
 import java.time.Duration;
 import java.time.Instant;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
@@ -89,6 +91,9 @@ import static net.kyori.adventure.text.format.TextDecoration.BOLD;
  * A collection of formatted messages used by the plugin.
  */
 public interface Message {
+
+    static final DateTimeFormatter DATE_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd '@' HH:mm:ss")
+            .withZone(ZoneId.systemDefault());
 
     TextComponent OPEN_BRACKET = Component.text('(');
     TextComponent CLOSE_BRACKET = Component.text(')');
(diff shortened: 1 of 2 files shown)


Diddyy:
@emilyy-dev Would you be able to review? Saw you tagged the issue a week ago. 
Any feedback would be appreciated 😅

common/src/main/java/me/lucko/luckperms/common/locale/Message.java
```diff
@@ -3714,6 +3719,7 @@ static TextComponent prefixed(ComponentLike component) {
                                     .color(GRAY)
                                     .key("luckperms.duration.since")
                                     .args(DurationFormatter.CONCISE_LOW_ACCURACY.format(action.getDurationSince()))
+                                    .hoverEvent(HoverEvent.showText(text().append(text("Date: ", GRAY)).append(text(DATE_FORMAT.format(action.getTimestamp()), AQUA))))
```
Aberdeener, line 3722: This `"Date:"` should be translatable and not hardcoded (see 2 lines above).
Translation terms live in `properties` files, for example the English one:
https://github.com/LuckPerms/LuckPerms/blob/master/common/src/main/resources/luckperms_en.properties

You can see some usage of the `translatable()` and `key()` methods in this class, it should be fairly simple to do!

common/src/main/java/me/lucko/luckperms/common/locale/Message.java
```diff
@@ -90,6 +92,9 @@
  */
 public interface Message {
 
+    static final DateTimeFormatter DATE_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd '@' HH:mm:ss")
```
Aberdeener, line 95: Wondering if this should be translatable/configurable? Many countries have vastly different time formatting standards

Diddyy: Reason I did it in that format is that’s what ‘ VerboseListener’ and ‘ExportCommand’ use for date format. But it would be cool to add 😅

common/src/main/java/me/lucko/luckperms/common/locale/Message.java
Diddyy: Good suggestion, I’ll change that soon!

common/src/main/java/me/lucko/luckperms/common/locale/Message.java
Aberdeener: Ah okay, no worries in that case then I think!

common/src/main/java/me/lucko/luckperms/common/locale/Message.java
lucko: Any update? :) happy to merge this once that change has been made

Diddyy: > Any update? :) happy to merge this once that change has been made

I have just commited the changes I believe. I haven't managed to check it just yet as I have no access to a server atm but I think it should be right..

lucko:
Thanks for the PR :)

PR #3606, opened 2023-03-19
Handle invalid empty command inputs

emilyy-dev:
Reject empty inputs for:
* Meta key
* Group display name
* Permission

Additionally, reject invalid group names in `lp group .. delete`.
Behaviour was also expanded to and documented in API `NodeBuilder`s.


emilyy-dev:
Oh whoops, the translatable won't work well with the text literal argument

emilyy-dev:
The problem with the translation is resolved, I simply added a new entry for each message, this is already done for other entries as well, and I don't think there are enough differences to really worry about separating them into their own translatable arguments, so I deemed this solution the least complex.

The issue with allowing spaces in inheritance nodes needs to be discussed further.

lucko (approved):

PR #3646, opened 2023-05-03
fix race

sunmisc:
when reading an object through a race, we are required to stretch two states - release / acquire

in this context, initialization does not go through a race, but through a synchronization block, but reading an object without synchronization is not valid in this case

two options:
add synchonized (blocking) to all methods everywhere - not the best option
and making the field volatile is a better option

But you can go further, creating CHM is a cheap operation, with low competition for this method it is better to do CAS in for-loopE, so we get much more performance

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/treeview/TreeNode.java
+++ b/common/src/main/java/me/lucko/luckperms/common/treeview/TreeNode.java
@@ -30,6 +30,7 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
@@ -59,7 +60,8 @@ public class TreeNode {
         }
     }
 
-    private Map<String, TreeNode> children = null;
+    private final AtomicReference<ConcurrentMap<String, TreeNode>>
+            children = new AtomicReference<>();
 
     private final int level;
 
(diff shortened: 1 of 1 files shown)


lucko:
Thanks, but this seems like a premature fix (we only ever insert into this map from a single thread)

sunmisc:
> Thanks, but this seems like a premature fix (we only ever insert into this map from a single thread)

just for reasons that you have getChildMap published through a race, I made changes, if this method is called in one thread, we don’t need a lock at all (synchonized)

otherwise, if getChildMap is still rummaged through different threads, then we still have to put volatile on the map (even if the other methods are read by 1 thread), the map is published through a race, you can also replace CHM with HashMap

common/src/main/java/me/lucko/luckperms/common/treeview/TreeNode.java
```diff
@@ -73,13 +75,13 @@ public TreeNode() {
...
-    private synchronized Map<String, TreeNode> getChildMap() {
-        if (this.children == null) {
-            this.children = new ConcurrentHashMap<>();
-        }
-        return this.children;
+    private Map<String, TreeNode> getChildMap() {
+        ConcurrentMap<String, TreeNode> a;
+        do {} while ((a = children.get()) == null &&
+                !children.compareAndSet(null,
+                        a = new ConcurrentHashMap<>()));
```
hamza-cskn, line 82: Is there any unaesthetic reason to use do while like this?

sunmisc: no, absolutely not, you can use a regular while/for loop

Also, starting with jdk 9+, you can do it without a loop at all using compareAndExchange, but we are using java 8

PR #3670, opened 2023-06-11
Add support for redis cluster

Dominik48N:
I once expanded the Redis support so that you can also use messaging with a Redis cluster.

I only made `addresses` from the `adress` in the config and a string list where you can specify the IP addresses and ports of your Redis nodes.

### DIFF
--- a/bukkit/src/main/resources/config.yml
+++ b/bukkit/src/main/resources/config.yml
@@ -261,11 +261,11 @@ push-log-entries: true
 broadcast-received-log-entries: true
 
 # Settings for Redis.
-# Port 6379 is used by default; paste the addresses of your nodes in the format "host:port" under addresses.
+# Port 6379 is used by default; set address to "host:port" if differs
+# Multiple Redis nodes can be specified in the same format as a string list under the name "addresses".
 redis:
   enabled: false
-  addresses:
-    - localhost
+  address: localhost
   username: ''
   password: ''
 
--- a/bukkit/src/main/resources/config.yml
+++ b/bukkit/src/main/resources/config.yml
@@ -264,7 +264,8 @@ broadcast-received-log-entries: true
 # Port 6379 is used by default; set address to "host:port" if differs
 redis:
   enabled: false
-  address: localhost
+  addresses:
+    - localhost
   username: ''
   password: ''
 
--- a/bukkit/src/main/resources/config.yml
+++ b/bukkit/src/main/resources/config.yml
@@ -261,7 +261,7 @@ push-log-entries: true
 broadcast-received-log-entries: true
 
 # Settings for Redis.
-# Port 6379 is used by default; set address to "host:port" if differs
+# Port 6379 is used by default; paste the addresses of your nodes in the format "host:port" under addresses.
 redis:
   enabled: false
   addresses:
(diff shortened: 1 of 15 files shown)


lucko:
I'm happy to merge this change in, but it needs to be backwards compatible please! (people's existing config files should keep working as they did before)

Dominik48N:
Do you think it makes more sense to create another config key for the multiple addresses or that I offer support for a string & a string list from the old config entry?

powercasgamer:
> Do you think it makes more sense to create another config key for the multiple addresses or that I offer support for a string & a string list from the old config entry?

Should check if `address` exists and use that (and possibly migrate that to `addresses`), if it doesn't exists use `addresses`

Dominik48N:
I changed something. Does that correspond to such an idea?

lucko:
Yep much better, can we keep the config how it was before? (most people I assume will only be using a single instance)

Dominik48N:
Do you mean that in the default config there is only `address` instead of `addresses` and that `addresses` then only says in the wiki that this is available?

lucko:
Yea :)

Dominik48N:
I'll do it for you right away. But I also write that in the comment above the section so that you can read it directly in the config that you can add multiple nodes.

Dominik48N:
Is now back with before!

lucko:
The unit/integration tests are failing, I think probably just an import that needs to be updated in the standalone module

Dominik48N:
The old Jedis version doesn't include some features that I used. I think it was because of that.

lucko:
That's fine, you'll also need to update this file https://github.com/LuckPerms/LuckPerms/blob/master/common/src/main/java/me/lucko/luckperms/common/dependencies/Dependency.java

and run the DependencyChecksumTest to figure out the new hash

Dominik48N:
I have adjusted.
I tested it on Velocity 3.2.0-SNAPSHOT and Paper 1.19.4 with two Redis Nodes (7.0.11).

Dominik48N:
I can not understand. My tests pass at my place?

Dominik48N:
I refresh my whole local project. I'm very confused right now, sorry.

lucko:
By default the docker integration tests are skipped, try commenting out this line

https://github.com/LuckPerms/LuckPerms/blob/master/standalone/src/test/java/me/lucko/luckperms/standalone/MessagingIntegrationTest.java#L56

and running the redis test :)

Dominik48N:
Sorry, I've been looking for the problem all this time.
I didn't consider that normal Redis instances logically don't work in a cluster.

Since the cluster support is also disabled in the test, the test (fortunately) does not pass. Since I tested it in a Redis cluster, I was always wondering why it works for me. I'll work on it and add additional support for it when cluster support is disabled.

lucko:
No problem - the tests did their job then! :)

Dominik48N:
The tests pass and now it also works for non-Redis cluster setups. I also tested it again myself, for single node setups, with the latest Redis Docker Version, Velocity 3.2.0-SNAPSHOT and Paper 1.19.4.

Please have a look again 👍

image

powercasgamer (changes requested):

powercasgamer (approved):

lucko:
Turns out there is a new version of `JedisPool` called `JedisPooled` which implements a common interface to `JedisCluster`, which makes things a bit tidier I think :)

I've re-ran the tests and everything looks good! Thanks for the PR :)

PR #3673, opened 2023-06-12
Improve "console" detection on fabric

DrexHD:
The fabric console detection currently simply checks if the `ServerCommandSource` has an entity "attached".
However, this causes two problems:

1. The execute command can attach entities to the command source, eg: `/execute as @a run lp user @s parent add player`. When this command is executed from the console, only players with permission to run `lp user @s parent add player` will receive the group. The version info message is printed into the console for any other player. Allowing modifiers like this allows for better datapack integration with luckperms commands. 
2. There is currently a small security issue. It is currently possible to run any luckperms commands as an OP player, by using a command block to execute them (because that doesn't have an entity attached), even though OP players can't execute luckperms commands by default.

This PR rewrites the fabric `isConsole()` check to use the `CommandOutput` instead. To explicitly check the command source against console, functions and rcon.


lucko:
Thanks for this :)

PR #3683, opened 2023-06-17
Fix that one bug and compatibility with my plugin (Allow enabling/disabling giveDefaultIfNeeded)

marlester-dev:
There was a small talk in luckperms discord about it, this PR should fix it. It allows API users to enable/disable giveDefaultIfNeeded which was the thing that prevented compatibility with my plugin, it will not affect API code of current API users, their code will stay ok.

### DIFF
--- a/api/src/main/java/net/luckperms/api/model/user/UserManager.java
+++ b/api/src/main/java/net/luckperms/api/model/user/UserManager.java
@@ -121,12 +121,12 @@ public interface UserManager {
      * <p>You should call this after you make any changes to a user.</p>
      *
      * @param user the user to save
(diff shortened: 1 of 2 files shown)


lucko:
What are you trying to do in your plugin?

lucko:
What storage backend are you using?

lucko:
I have committed a fix for the underlying issue & validated with unit tests that LP is now behaving correctly. This workaround should not be necessary :)

https://github.com/LuckPerms/LuckPerms/commit/9f1e74fa7c5ce8226d92a9ab02aa383111c6c461

PR #3748, opened 2023-09-05
Fixed the 1.20.1 respawn bug with capabilities issue. Fixes #3747

### LINKED ISSUE #3747: Death Loop Bug on 1.20.1
### Description

When a player dies and any mod listens to both luck perms and the death event it causes players to not be able to respawn.,

### Reproduction Steps

1. Have a mod that does something on death and interacts with luckperms
2. Die  in game
3. Can't respawn correctly
4. Delete player data to respawn

luck perms

### Expected Behaviour

Players to respond normally

### Server Details

Minecraft Forge 1.20.1-47.1.3

### LuckPerms Version

5.4.102

### Logs and Configs

[Relevant Part](https://mclo.gs/ed0VRXL)
[Full Log](https://mclo.gs/MYrHLML)

### Extra Details

_No response_

NullOrNaN:
This request fixes the issues I was having and seeing extensively and causing problems with my players.  We've had to spend more time than I care to admit to debug and restore inventories and player data to resolve this issue. 

As a result of the headache, we've gone ahead and found what was causing it and issued a local fix ourselves, then forked the repo and submitted the patch to you.  Hopefully, this prevents others from having this issue and our headache as well.

### DIFF
--- a/forge/src/main/java/me/lucko/luckperms/forge/capabilities/UserCapabilityImpl.java
+++ b/forge/src/main/java/me/lucko/luckperms/forge/capabilities/UserCapabilityImpl.java
@@ -63,6 +63,9 @@ public class UserCapabilityImpl implements UserCapability {
      * @return the capability
      */
     public static @NotNull UserCapabilityImpl get(@NotNull Player player) {
+        if (!player.getCapability(CAPABILITY).isPresent()) {
+            player.reviveCaps();
+        }
         return (UserCapabilityImpl) getCapability(player).orElseThrow(() -> new IllegalStateException("Capability missing for " + player.getUUID()));
     }
 


Brokkonaut:
thank you :) we we had the same issue on our server

NullOrNaN:
> thank you :) we we had the same issue on our server

You're welcome. I'm hopeful that it will help many people and that LuckPerms will accept the PR due to the nature of the bug.

lucko:
The reason I haven't merged this yet is because I'm not sure whether this is really the correct fix. I think we need to be careful about "revivingCaps" when the player genuinely is in an inactive state. It feels to me like Forge need to fix the capabilities system at their end if it is causing problems. As far as I can tell, LuckPerms' usage of the system is correct.

Maybe we should just switch to a centralised `Map<UUID, User>` instead of using the capabilities if it is going to continue to be painful

NullOrNaN:
I agree Forge is to blame. We sadly have to work around them sometimes due to the nature of that project. Most will likely move to NeoForge, and I have higher hopes for them. For now, while it's not the ideal fix, I didn't want to re-work your entire plugin for any platform this touches. 

Changing it to a map might be better; however, if we go that route vs this patch, you'd have to backport it to 1.20. and 1.20.1 vs. forward porting. The choice is ultimately up to you.

Icosider:
Hello, i was be testing this problem on my main servers and trying fix error in this way:

UserCapabilityListener
```java
private final ThreadLocal<Player> previousPlayerData = new ThreadLocal<>();

@SubscribeEvent
public void onPlayerClone(PlayerEvent.Clone event) {
    Player previousPlayer = event.getOriginal();
    Player currentPlayer = event.getEntity();

    previousPlayer.reviveCaps();
    try {
        UserCapabilityImpl previous = UserCapabilityImpl.get(previousPlayer);
        UserCapabilityImpl current = UserCapabilityImpl.get(currentPlayer);

        current.initialise(previous);
        current.getQueryOptionsCache().invalidate();
    } catch (IllegalStateException e) {
        // continue on if we cannot copy original data
    }
    previousPlayerData.set(previousPlayer);
}

@SubscribeEvent
public void onPlayerRespawn(PlayerEvent.PlayerRespawnEvent event) {
    if (previousPlayerData.get() != null) {
        previousPlayerData.get().invalidateCaps();
    }
}
```

UserCapabilityImpl:
Remove the getCapability(Player): LazyOptional<UserCapability> method and make this:
```java
public static @NotNull UserCapabilityImpl get(@NotNull Player player) {
    return (UserCapabilityImpl) player.getCapability(CAPABILITY).orElseThrow(() -> new IllegalStateException("Capability missing for " + player.getUUID()));
}

public static @Nullable UserCapabilityImpl getNullable(@NotNull Player player) {
    return (UserCapabilityImpl) player.getCapability(CAPABILITY).resolve().orElse(null);
}
```

Now let me explain why I did it that way. First, when the player respawns, Capabilities are attached, after which the player is cloned. At the end of all this action, permissions are sent to the client. I noticed that with previous fixes, you tried to get data from the player's Capability using reviveCaps and then invalidateCaps. But you also do this in UserCapabilityImpl#getCapability, inside which you also get Capability with recovery, and then invalidateCaps. Thus, the Capability of your player (the expected one) becomes invalid already at the stage of its creation (in our case respawn). I removed all the extra code and divided Clone into some 2 phases: Clone - copying data and restoring old ones, and PlayerRespawn, when the player is already fully ready to appear in the world and the need for old data is not needed.

Perhaps ThreadLocal will be superfluous in my version, or maybe not, I did not check it thoroughly. But in the current variation, there is no error at entry or after death.

If many people's mistake is corrected due to my extraordinary decision and @lucko himself or someone from the LuckPerms team gives a comment or consent to the implementation, then I will be ready to provide a PR for this code

lucko:
A similar fix was added in https://github.com/LuckPerms/LuckPerms/commit/5c1ea5633ed5986c02d3fa1ccf187638d143c091 which should fix the problem :)

PR #3777, opened 2023-11-06
Fix inability to correctly delete conflicting uuid

huanmeng-qwq:


common/src/main/java/me/lucko/luckperms/common/storage/implementation/sql/SqlStorage.java
```diff
@@ -58,28 +58,15 @@
...
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
underscore11code, line 63: Imports should be left explicit, not wildcards.

common/src/main/java/me/lucko/luckperms/common/storage/implementation/sql/SqlStorage.java
```diff
@@ -58,28 +58,15 @@
...
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
 public class SqlStorage implements StorageImplementation {
-    private static final Type LIST_STRING_TYPE = new TypeToken<List<String>>(){}.getType();
+    private static final Type LIST_STRING_TYPE = new TypeToken<List<String>>() {
+    }.getType();
```
underscore11code, line 69: Not sure why this is changed?

common/src/main/java/me/lucko/luckperms/common/storage/implementation/sql/SqlStorage.java
```diff
@@ -129,7 +116,7 @@ public class SqlStorage implements StorageImplementation {
     private static final String ACTION_SELECT_ALL = "SELECT * FROM '{prefix}actions'";
 
     private final LuckPermsPlugin plugin;
-    
+
```
underscore11code, line 119: Or this?

underscore11code (changes requested):
You've made some code style changes I suspect Luck won't like, I'd recommend reverting them to improve the chances of Luck approving.

huanmeng-qwq:
Sorry, this is due to my IDEA automatically performing formatting

lucko:
Hey, thanks for the PR :)

I have reverted the changes you made to `SqlStorage.java` - I think maybe you misunderstood what the SQL statement was doing:

```sql
DELETE FROM '{prefix}players' WHERE username=? AND NOT uuid=?
```

This has the effect of deleting all conflicts, which is the intention - so there's no need to make batch updates. However, maybe I have misunderstood, and if so, please feel free to reclarify what that change was intending to achieve.

I've also added some more tests to validate that this doesn't break again in the future :)

Thanks again for the PR!

PR #3822, opened 2024-01-22
Update mysql.sql

rushilhacker1:


emilyy-dev:
What's the reason behind this change? `BOOL`/`BOOLEAN` both mean the same thing.
And the `luckperms_` table prefix should not be hardcoded, that's what the `{prefix}` placeholder is for, as it's configurable in the plugin settings.

rushilhacker1:
i had to run the schema on my own so the bool boolean gave error and similarly prefix

PR #3836, opened 2024-02-15
Don’t read permission nodes twice for users

FixedDev:
After analyzing the code on the SqlStorage, I found that line of code when selecting user permission nodes.

After checking again, I came to the conclusion that reading the node again was unnecessary and wouldn’t affect the behavior.


lucko:
Ah oops! Nice spot, thanks :)

PR #3863, opened 2024-03-25
[Fabric] Fix players getting kicked on long taking configuration and make setting the user run earlier

Patbox:
First commit fixes the issues when mods try to check permissions which could happen before luckperm's event could be executed.
Second commit fixes issue where players that took too long to join (for example because of using big resource pack on slow connection), got kicked out because of luckperms

### DIFF
--- a/fabric/src/main/java/me/lucko/luckperms/fabric/mixin/ServerConfigurationNetworkHandlerMixin.java
+++ b/fabric/src/main/java/me/lucko/luckperms/fabric/mixin/ServerConfigurationNetworkHandlerMixin.java
@@ -26,7 +26,9 @@
 package me.lucko.luckperms.fabric.mixin;
 
 import com.llamalad7.mixinextras.sugar.Local;
+import com.mojang.authlib.GameProfile;
 import me.lucko.luckperms.fabric.event.PreOnPlayerConnectCallback;
+import me.lucko.luckperms.fabric.event.ServerConfigurationTickCallback;
 import net.minecraft.network.ClientConnection;
 import net.minecraft.network.packet.c2s.config.ReadyC2SPacket;
 import net.minecraft.server.MinecraftServer;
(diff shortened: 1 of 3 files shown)


fabric/src/main/java/me/lucko/luckperms/fabric/mixin/ServerConfigurationNetworkHandlerMixin.java
```diff
@@ -0,0 +1,66 @@
...
+@Mixin(ServerConfigurationNetworkHandler.class)
+public abstract class ServerConfigurationNetworkHandlerMixin extends ServerCommonNetworkHandler {
+    @Shadow @Final private GameProfile profile;
+
+    public ServerConfigurationNetworkHandlerMixin(MinecraftServer server, ClientConnection connection, ConnectedClientData clientData) {
+        super(server, connection, clientData);
+    }
+
+    @Inject(method = "onReady", at = @At(value = "INVOKE", target = "Lnet/minecraft/server/PlayerManager;onPlayerConnect(Lnet/minecraft/network/ClientConnection;Lnet/minecraft/server/network/ServerPlayerEntity;Lnet/minecraft/server/network/ConnectedClientData;)V", shift = At.Shift.BEFORE), cancellable = true)
+    private void beforeConnecttion(ReadyC2SPacket packet, CallbackInfo ci, @Local ServerPlayerEntity player) {
```
lucko, line 55: Is there not a way to do this with Fabric API? :(

I would prefer to now have to maintain our own event for something as critical (to LuckPerms) as the login event

Patbox: All fabric events happen too late from my testing. This is the earliest point ServerPlayerEntity instance exists and it runs just before rest of server's initialization code does, which should cover all cases (as mods are also likely to check permissions with mixin at any point within the onPlayerConnect). Fabric is just a mixin-heavy platform, so there is no real way to avoid it.

DrexHD:
Bump! I would love to see this merged anyways. This ("permission checks not being ready") is very annoying, because it causes prefixes of joining players to be inconsistent!

lucko:
I feel quite strongly that we should use the builtin login events - and if these are not suitable (for what I would've thought is a pretty common usecase), then we should endeavour to PR improvements to Fabric API.

Neovitalism:
Honestly, you don't need the event, just the mixin. When it comes to how the event is too specific, using the mixin without creating your own event is beneficial as nobody else will use it lol

For the sake of servers that have this issue, highly suggest this fix. I've seen plenty of servers get the issue over the course of the year, and something is definitely needed there.

PR #3881, opened 2024-04-28
Set mappings namespace to skip remapping on Paper 1.20.5+

jpenilla:
Remap is not needed since we don't use any non-reflection NMS (or NMS in general, afaik).

### DIFF
--- a/bukkit/loader/build.gradle
+++ b/bukkit/loader/build.gradle
@@ -25,6 +25,10 @@ shadowJar {
     from {
         project(':bukkit').tasks.shadowJar.archiveFile
     }
+
+    manifest {
+        attributes(["paperweight-mappings-namespace": "mojang"])
+    }
 }
 
 artifacts {


bukkit/build.gradle
```diff
@@ -48,6 +48,10 @@ shadowJar {
     relocate 'com.rabbitmq', 'me.lucko.luckperms.lib.rabbitmq'
     relocate 'org.apache.commons.pool2', 'me.lucko.luckperms.lib.commonspool2'
     relocate 'ninja.leaping.configurate', 'me.lucko.luckperms.lib.configurate'
+
+    manifest {
+        attributes(["paperweight-mappings-namespace": "mojang"])
+    }
```
lucko, line 54: I think this probably needs to be set on the 'outer' jar as I assume that's where Paper looks, i.e. here: https://github.com/LuckPerms/LuckPerms/blob/master/bukkit/loader/build.gradle

jpenilla: Oops, should be fixed now

PR #3883, opened 2024-05-05
Update Forge to MC 1.20.6

bloodmc:
This PR addresses the following

- Updates Forge branch to MC 1.20.6
- Fixes user capability initialization [issue](https://gist.github.com/bloodmc/467dd7b78c384825588f06b6301c1f87) on player first login.
- Changes shadow plugin to updated [fork](https://github.com/Goooler/shadow) to fix shadow bugs with Gradle 8.6/JDK 21+. See https://github.com/johnrengelman/shadow/pull/876
Note: Original [shadow repo](https://github.com/johnrengelman/shadow) has been dead for over a year.


IMB11:
Should really be supporting NeoForge at this point, most the playerbase has moved to that for 1.20.1+

forge/src/main/java/me/lucko/luckperms/forge/util/BrigadierInjector.java
```diff
@@ -139,7 +149,26 @@ private InjectedPermissionRequirement(String permission, Predicate<CommandSource
...
+                        Component component = TranslationManager.render(Message.LOADING_STATE_ERROR.build(), player.getLanguage());
+                        if (this.plugin.getConfiguration().get(ConfigKeys.CANCEL_FAILED_LOGINS)) {
+                            player.connection.disconnect(ForgeSenderFactory.toNativeText(component));
+                            return false;
+                        }
+                    }
+                    // initialise capability
+                    userCapability.initialise(user, player, this.plugin.getContextManager());
+                    this.plugin.getContextManager().signalContextUpdate(player);
+                }
```
lucko, line 171: It looks like this has just been copy pasted from

```java
@SubscribeEvent(priority = EventPriority.HIGHEST)
public void onPlayerLoadFromFile(PlayerEvent.LoadFromFile event) {}
```

in ForgeConnectionListener.java.

Maybe we are using the wrong event there? 

Other platforms have a PlayerLoginEvent or similar which is called when the player instance is first created - does Forge have an equivalent? It looks like `PlayerEvent.LoadFromFile` was chosen originally but I am not sure if that was the correct choice.

bloodmc: I dug in deeper and was able to clean this up a bit. The original issue here is that Forge will check command permissions before our user capability was initialised. However, rather than attempting to initialise, I now grab the user directly from LP if we are still in the `pre-login` state. I also changed the `PlayerEvent.LoadFromFile` to `PlayerLoggedInEvent` as it would not always be called on login. Everything seems to work as expected. 

Only question I have now is, should vanilla OP'd users have full access to LP commands? Because currently they do not but I recall other platforms working this way. It does not seem to be related to my changes.

lucko (approved):

PR #3931, opened 2024-06-30
Fix QueryOptionsCache referencing outdated player instance

DrexHD:
Respawning creates a new ServerPlayerEntity instance. The current code copies the QueryOptionsCache from the old player instance, however this causes an issue, because the cache still holds the reference to the old entity. So any usage of it will invoke ContextManager.calculate causing outdated context to be calculated. This is simply fixed by creating a new QueryOptionsCache with the new instance.
TLDR: Currently context changes are no longer updated (on fabric) after player deaths, this PR fixes this.


OGBubba22:
I believe this may fix the current issue I am having. Is there anyway you can dumb it down and explain how to fix this/how to create a new QueryOptionsCache? 
[New_Text_Document.txt](https://github.com/user-attachments/files/16103390/New_Text_Document.txt)

lucko:
Thanks for the PR :)

PR #3950, opened 2024-08-05
Port to NeoForge

jpenilla:
Initial port to NeoForge. There are a couple of TODOs left and it could use some general review and testing. But has been tested to get into world with commands working on the dedicated and integrated server.


lucko:
Hey @jpenilla - thank you for this!

I have thought a bit more, I think we should keep supporting regular Forge for the time being in addition to NeoForge. If it becomes too much of a burden, Forge will be removed.

Would you be able to re-instate the Forge module please? :)

jpenilla:
@lucko I've added back the forge module. 
Do you have any thoughts on requiring an IModInfo or ModContainer for event registration?
https://github.com/LuckPerms/LuckPerms/pull/3950/files#diff-ebe2107a53d10525584eceb9c3d0ecac119f690a38351288dd255307f6088b1aR41

lucko:
I'm unsure - are there any places in NeoForge where you have to pass either of these? It should be a class similar to Bukkit's `Plugin` interface.

Another way to look at it - which is easier to get hold of if you're writing a mod?

jpenilla:
Looks like `ModContainer` is used much more often.
Mod authors can get their `ModContainer` by declaring it as a parameter of their `@Mod` class constructor:
https://github.com/LuckPerms/LuckPerms/pull/3950/files#diff-8f07a3e4f1d64aac49ffd7402fba4eb0e5028ee62fedc1f5bfd5f91101caf111R52

The `ModContainer` seems to be more analogous to a `Plugin` as well, with `IModInfo` being more like the plugin description. Extension points, configs, etc are registered through it.

lucko:
Thanks very much for this :)

PR #3961, opened 2024-08-21
Fix login configuration phase being blocked by un-ACK-ed configuration task on NeoForge

### LINKED ISSUE #3960: Client hangs when joining a NeoForge server with just LuckPerms installed
### Description

Client will stuck at "Joining World...." when joining a Minecraft 1.21.1 server with just NeoForge and LuckPerms installed.

### Reproduction Steps

1. Setup a Minecraft 1.21.1 server
2. Install NeoForge 21.1.8
3. Get & install `LuckPerms-NeoForge-5.4.139.jar`
4. Start the server 
5. Get a Minecraft 1.21.1 client. It can be either vanilla or w/ NeoForge.
6. Join the server in step 4

### Expected Behaviour

You join in the game.

### Server Details

Minecraft 1.21.1 w/ NeoForge 21.1.8

### LuckPerms Version

5.4.139

### Logs and Configs

There is no error in either server or client log.

Nevertheless, this is the `debug.log` from server, with logging level manually set to `TRACE`.

https://gist.github.com/3TUSK/d3f414d34abbe193c20cc20f2f8a9dc7

### Extra Details

It was only mentioned here https://github.com/LuckPerms/LuckPerms/issues/3936#issuecomment-2287836735, and I managed to let it slip away in my first few rounds of search. 

Considering the nature of this issue, I think it is worth having a new ticket.

3TUSK:
According to NeoForge Documentation, section "Networking § Using Configuration Tasks":

> When the configuration is not acknowledged, then the server will wait forever, and the client will never join the game. So it is important to always acknowledge the configuration task, unless the configuration task failed, then you can disconnect the client.

As of now, LuckPerms never marks `USER_LOGIN_TASK_TYPE` as completed even if it actually finishes. This leads to client hanging on login, as server was mistakenly waiting for a acknowledgement from client which never happens.

This PR slightly modifies the `AsyncConfigurationTask` so that it will "acknowledge" the configuration task on its own, as described in [NeoForge's own documentation](https://docs.neoforged.net/docs/networking/configuration-tasks#acknowledging-a-configuration-task).

This closes #3960.

### DIFF
--- a/neoforge/src/main/java/me/lucko/luckperms/neoforge/listeners/NeoForgeConnectionListener.java
+++ b/neoforge/src/main/java/me/lucko/luckperms/neoforge/listeners/NeoForgeConnectionListener.java
@@ -81,7 +81,7 @@ public class NeoForgeConnectionListener extends AbstractConnectionListener {
 
         event.register(new AsyncConfigurationTask(this.plugin, USER_LOGIN_TASK_TYPE, () -> CompletableFuture.runAsync(() -> {
             onPlayerNegotiationAsync(event.getListener().getConnection(), uniqueId, username);
-        }, this.plugin.getBootstrap().getScheduler().async())));
+        }, this.plugin.getBootstrap().getScheduler().async()), event.getListener()));
     }
 
     private void onPlayerNegotiationAsync(Connection connection, UUID uniqueId, String username) {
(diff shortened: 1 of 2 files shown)


lucko:
Thanks for the PR!

(I amended it slightly, and since you opened the PR from an organisation I couldn't push to your branch - so had to merge it manually)

PR #3969, opened 2024-09-08
Simplify ExpiringSet implementation

AlexProgrammerDE:
Switches the custom ExpiringSet implementation to use a java-backed implementation from Collections with more methods implemented.

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/util/ExpiringSet.java
+++ b/common/src/main/java/me/lucko/luckperms/common/util/ExpiringSet.java
@@ -25,36 +25,21 @@
 
 package me.lucko.luckperms.common.util;
 
-import com.github.benmanes.caffeine.cache.Cache;
-
+import java.util.Collections;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
-/**
- * A simple expiring set implementation using Caffeine caches
- *
- * @param <E> element type
- */
-public class ExpiringSet<E> {
-    private final Cache<E, Long> cache;
-    private final long lifetime;
+public final class ExpiringSet {
+    private ExpiringSet() {}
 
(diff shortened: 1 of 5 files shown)


common/src/main/java/me/lucko/luckperms/common/util/ExpiringSet.java
```diff
@@ -25,36 +25,21 @@
...
-    public boolean add(E item) {
-        boolean present = contains(item);
-        this.cache.put(item, System.currentTimeMillis() + this.lifetime);
-        return !present;
-    }
-
-    public boolean contains(E item) {
-        Long timeout = this.cache.getIfPresent(item);
-        return timeout != null && timeout > System.currentTimeMillis();
-    }
```
lucko, line 55: I think Caffeine `expireAfterWrite` doesn't guarantee that read calls will always omit expired entries, hence the need to check the timeout separately as well - maybe this has changed recently, can you confirm?

AlexProgrammerDE: https://www.javadoc.io/doc/com.github.ben-manes.caffeine/caffeine/2.2.0/com/github/benmanes/caffeine/cache/Caffeine.html#expireAfterWrite-long-java.util.concurrent.TimeUnit-

> Expired entries may be counted in [Cache.estimatedSize()](https://www.javadoc.io/static/com.github.ben-manes.caffeine/caffeine/2.2.0/com/github/benmanes/caffeine/cache/Cache.html#estimatedSize--), but will never be visible to read or write operations. Expired entries are cleaned up as part of the routine maintenance described in the class javadoc.

As I interpret it, expired entries are never visible in read and write operations.

lucko:
Thanks for the PR :)

PR #4018, opened 2025-01-07
Add fabric console permission bypass

DrexHD:
Currently consoles on fabric (and presumably forge, neoforge) can't execute commands that require a permission and don't use an operator fallback.

This is an initial draft for a console bypass *(console always has all permissions)* on fabric. I am not sure on the specifics of the verbose check event, let me know what you think! If this approach seems fine, it should be easy to port these changes to forge and neoforge.


DrexHD:
Sender factories need to be initialized earlier, to prevent NPE when datapack (and similar) load and check permissions, with this change.

PR #4040, opened 2025-03-09
Allow specifying contexts on import

Kas-tle:
This PR allows specifying a context to apply when importing an export file or web export. This is useful when doing multiple imports from previously separate servers that will now be accessed through a single proxy.


Kas-tle:
On second thought, I think I will need to modify this further in order to only merge nodes when the keys on the existing node are of the single and same type as the imported node.

PR #4041, opened 2025-03-09
Config option to combine existing nodes on import with new nodes of same key, value, & expiry

Kas-tle:
Related to https://github.com/LuckPerms/LuckPerms/pull/4040

Adds a config option (default `false`) to merge existing node entries when using the import command. This results in a much cleaner permissions set when there are overlapping nodes due to different contexts.

Usage:
```yaml
# If LuckPerms should attempt to merge into existing permission nodes when importing data when the 
# effective result would be the same.
#
# - Useful if you are importing data from multiple sources with overlapping permissions.
# - Note that this will cause the import process to be slower.
use-existing-node-entries-on-import: false
```


Kas-tle:
On second thought, I think I will need to modify this further in order to only merge nodes when the keys on the existing node are of the single and same type as the imported node.

Issue #4049, opened 2025-03-19
World context not functioning correctly after relog.


andrewkm:
### Description

We have been battling with weird permissions problems for ages now due to world contexts, and just finally tracked it down.
Basically; after relogging to the server, world context is no longer obeyed correctly. (Some sort of memory issue perhaps?)

After a server restart, on the first login of a player, reproduction is impossible no matter how much you go at it. Functionality is correct. However, after the player disconnects and reconnects to the server, reproduction is instant.

### Reproduction Steps

Here is a full start to stop server demonstration.

The setup is as follows;
Only Luckperms & Multiverse on the server, as Multiverse needed for world switching.
We have added the permission multiverse.teleport.* as true in server=main and world=rising_s
(Do note this works with any permission from any plugin, I merely selected one at random).
We are dealing with two worlds, rising_s and mining_s.
We have enabled verbose to observe LuckPerms.

**Immediately** after a server restart we switch worlds around by throwing the player around via console using Multiverse's /mvtp command and can observe that the players permission correctly works in rising_s and doesn't work in mining_s.

https://mclo.gs/YmXqtk5
**Lines** 240, 276, 312 show permission being correctly **granted** on rising_s world.
**Lines** 244, 280 show  permission being correctly **denied** on mining_s world.

**Line** 315-319 show a relog to the server.

**Lines** 324, 332 show permission being **INCORRECTLY** granted on mining_s world.

A server restart, once again fixes issues, until the player relogs and is able to once more break the system.
There is something going on here / being stuck in memory, cache, or some funky-ness along those lines.

Hopefully it's not too difficult of a fix.

### Expected Behaviour

Permissions to be correctly granted and denied on world change, post relogging to the server.

### Server Details

Paper 1.21.4-211-main@6ea4202

### LuckPerms Version

LuckPerms v5.4.156

### Extra Details

If you need any additional details please let me know.

BNTFryingPan:
please check (and share) the output of `/lp user <name> info` when each of these checks occurs. i am unable to reproduce this on a test server. the world context updates instantly, and relogging does not change its value

andrewkm:
> please check (and share) the output of `/lp user <name> info` when each of these checks occurs. i am unable to reproduce this on a test server. the world context updates instantly, and relogging does not change its value

https://mclo.gs/GGaI2lo
Line 415 is the incorrect granting of a permission that should not have been granted.

Prior to the disconnection on Line 388 all was working correctly.
[...]

PR #4050, opened 2025-03-19
Improve context manager caching

lucko:
Maybe a fix for #4049

In the past for some platforms, LuckPerms used a global map as a fallback for locating `QueryOptionsCache` instances - this has proven to be a bit brittle and potentially the cause of some memory leaks, although never confirmed.

I can't see anything specifically wrong in the old code, but using `Player` as the key in a map feels wrong and I have a feeling it is the cause of the issues observed in #4049.

### DIFF
--- a/bukkit/src/main/java/me/lucko/luckperms/bukkit/context/BukkitContextManager.java
+++ b/bukkit/src/main/java/me/lucko/luckperms/bukkit/context/BukkitContextManager.java
@@ -25,45 +25,25 @@
 
 package me.lucko.luckperms.bukkit.context;
 
-import com.github.benmanes.caffeine.cache.LoadingCache;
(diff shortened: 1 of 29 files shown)


lucko:
[LuckPerms-Bukkit-5.4.157.jar.zip](https://github.com/user-attachments/files/19349449/LuckPerms-Bukkit-5.4.157.jar.zip)

andrewkm:
> [LuckPerms-Bukkit-5.4.157.jar.zip](https://github.com/user-attachments/files/19349449/LuckPerms-Bukkit-5.4.157.jar.zip)

Initial tests looking good here.
Going to play with it some more then toss it into production.
Thank you!

The amount of issues this fixes is wild!
Constant weirdness through out the years lol.

PR #4062, opened 2025-04-06
IPV6 

SonnenPingu:
Update HikariConnectionFactory.java for IPv6 support.

Should be able to recognise both.

Just testing it on my server 

Work on ipv6 ;)


SonnenPingu:
[10:10:57 INFO]: [LP] An update task has been requested, please wait.
[10:10:58 INFO]: [LP] Update task complete. Now attempting to push to other servers.
[10:10:58 INFO]: [LuckPerms] [Messaging] Sending ping with id: a016ba9c-2dd2-4267

[10:10:58 INFO]: [LP] Other servers were notified via Sql Messaging successfull

SonnenPingu:
I have only tested this with Bukkit/Paper so far. As I don't have any other server types.
So it would be good if you test this on the other platforms first :D

PR #4107, opened 2025-06-10
Straightforward Library for Custom Platforms

### LINKED ISSUE #4105: Straightforward Library Usage for Custom Platforms
### Description

I'm attempting to use LuckPerms as a permissions manager for a (Minecraft-related) standalone program. Right now, there is a *lot* of boilerplate (see the amount of code for [LuckPerms/standalone](https://github.com/LuckPerms/LuckPerms/tree/master/standalone)).

### Proposed Behaviour

Ideally, I would be able to do something like this after adding LuckPerms as a Maven dependency:
```
// StandaloneLuckPerms(String platformName, LuckPermsConfig config, Path dataFolder, Consumer<String> consoleOutput, BiConsumer<UUID, String> playerOutput)
// The config could also just be a ConfigurationLoader like in the ConfigurateConfigAdapter to make this suggestion easier to implement
StandaloneLuckPerms lp = new StandaloneLuckPerms("Platform Name", config, Path.of("luckperms_data_folder"), System.out::println, (playerUuid, msg) -> ...);
lp.start();

lp.execFromConsole(List.of("user", "mega12345mega", "permission", "set", "*", "true")); // Commands would come from my own CLI
lp.execFromPlayer(playerUuid, List.of(...));

Tristate permTrisate = lp.getPermissionState(playerUuid, "permission.name");
boolean hasPerm = lp.hasPermission(playerUuid, "permission.name");
assert hasPerm == permTristate.asBoolean()

lp.close();
```
My current messy code that does essentially this is ~600 lines. Note that I also am currently relying on [LuckPerms/common](https://github.com/LuckPerms/LuckPerms/tree/master/common) which itself could be made more straightforward by adding it to Maven: https://github.com/LuckPerms/LuckPerms/issues/3825

In my situation, I would also benefit from being able to embed all of the LuckPerms information in the SQLite database I'm already using instead of a data folder, so everything would be self-contained, but this isn't required. (Obviously this would significantly complicate things, especially considering the database will also have my own data in it)

### Extra Details

Right now, the dependency management is also a bit messy for this use case. Ideally, when I depend on this, all of the other dependencies would come with it and be managed by Maven instead of having to be loaded separately by LuckPerms at runtime and stored in a separate folder.

Context: Why use LuckPerms for something that isn't a Minecraft server? (This is still a server that players will be able to login to)
* Minecraft players are likely to already be familiar with LuckPerms
* LuckPerms already provides a group system, wildcards, web editor, and handles usernames and uuids out of the box

mega12345mega:
Closes #4105

This subproject will still need to be added to Maven for ease of use.

See [LuckPermsLibraryTest](https://github.com/mega12345mega/LuckPerms-library/blob/library/library/src/test/java/library/LuckPermsLibraryTest.java) for example usage. I'm not yet completely sure the best way to handle the dependencies - what do you recommend? Also, what are your thoughts on allowing the library user to supply more powerful storage options such that they would be able to embed everything in a single file (like mentioned in the issue) instead of just being able to specify the data folder path?

Before merging, it probably makes the most sense to modify [LuckPerms/standalone](https://github.com/LuckPerms/LuckPerms/tree/master/standalone) to just depend on this, which would cut out a lot of duplicated code (I made this by copying the standalone folder and just modifying some stuff). By design, it should be able to completely recreate the behavior of the standalone project (and it should only need a small file similar to LuckPermsLibraryTest).


mega12345mega:
The main difference between what I was suggesting doing and just refactoring the standalone module is that what I was suggesting involves adding a module which would allow you to download only the relevant code without anything specific for the standalone tool (such as the docker and terminal stuff). Otherwise, the actual amount of code would be basically equivalent. Is there any specific reason you would want to avoid adding a new module (even though the total amount of code in the repo is very similar)?

I hadn't considered using the REST API and standalone tool together, but this seems like a decent solution. However, this wouldn't have allowed for the changes I was considering for the storage format (which I would still like to add). If I'm reading the documentation correctly, it also looks like the REST API doesn't support executing commands in the same way; the `/action` has the target separated out rather than embedded into the command (I can't just say "execute from the perspective of \<user>: /lp user \<username> permission set \<permission> \<value>"; it wants "execute from the perspective of \<user> modifying \<username>: permission set \<permission> \<value>" which requires extra parsing before sending  the command).

Using the REST API like that also feels a bit convoluted; why have the program (1) unpack standalone LuckPerms from my program's resources, (2) start up a separate process, (3) then interact with that process through a REST API, and finally (4) make sure the LuckPerm files get properly cleaned up on program exit, when it can just execute the code directly like how Minecraft servers do?

The biggest reason to use LuckPerms over something like LDAP is familiarity, both for the users of my program and myself. It is also already designed for almost my exact use case, only, instead of a "Minecraft server," it's a "Minecraft-related server that still involves users logging in with their Minecraft accounts."

Note that my program is meant to be ran by anyone, not just ran on my system, which is why it should be self-contained and easy to setup.

BNTFryingPan:
> However, this wouldn't have allowed for the changes I was considering for the storage format

im pretty sure you can make an extension that adds a custom storage method already, and it would work on all platforms

mega12345mega:
I am now revisiting this pull request and my requirements have changed slightly, so I no longer need the ability to store everything within a single file. This greatly simplifies things, and was the main part that I had not yet completed. I believe the changes I have already made are similar to what I now need, and would like to finish up this pull request if possible. Now that I have been able to test this system in conjunction with my use case, I can confirm that this pull request is extremely useful for projects like mine.

@lucko
Therefore, I would like to check if this is something you think I should clean up and get merged, or if you have decided that supporting this is not a good idea and the pull request should simply be closed.

Note that I do not plan to maintain this personally, so if this does not get merged, I will simply archive the fork in case someone else wants to look at it. I have some alternatives that mean I do not *need* this for my project, but the alternatives are significantly worse than using LuckPerms in my opinion.

Issue #4135, opened 2025-07-19
NoClassDefFoundError when detecting Brigadier support on Java 17 using Paper 1.8.8 forks


MindsMaster:
### Description

When running LuckPerms v5.5.9 on a Paper 1.8.8 fork (e.g., ImanitySpigot) with Java 17+, the plugin fails to load due to a `NoClassDefFoundError` when calling `Class.forName(...)` to detect Brigadier support.

This issue does not occur under Java 8.

### Reproduction Steps

1. Use ImanitySpigot or another Paper 1.8.8-based fork
2. Run server with Java 17+
3. Install LuckPerms latest
4. Start the server

### Expected Behaviour

The plugin should detect that Brigadier is not present and continue loading normally, just like under Java 8.

### Server Details

imanityspigot(MC:1.8.8)

### LuckPerms Version

v5.5

### Logs and Configs

https://gist.github.com/MindsMaster/ec4a5e7096953cb262b093fb95142042

PR #4136, opened 2025-07-19
fix: Fix classExists to handle NoClassDefFoundError in Java 17

MindsMaster:
issue #4135


lucko:
thanks!

MindsMaster:
Sorry, I need to retry and reproduce the error I encountered at that time. I looked through the official documentation for Class.forName, and it doesn’t seem like it should throw a NoClassDefFoundError. However, I also wasn’t able to reproduce the issue just now. I’ll need some more time to re-examine this problem. Apologies for the disturbance, and thank you.

PR #4213, opened 2026-01-15
Hytale

lucko:
> [!CAUTION]
> This is a early preview and should be considered experimental.

An early build of LuckPerms for Hytale 🎉 

* **DOWNLOAD**: https://www.curseforge.com/hytale/mods/luckperms

I've done some brief testing and it mostly works! There are some things that don't work just yet, but will be updated in the coming days/weeks:

- [x] Refine hooks into Hytale's built-in permission system - it currently works but is a bit rough around the edges.
- [ ] Running a command via `/lp verbose command <command>` does not work
- [x] Integrate with Hytale API for username -> UUID lookups - you can only use usernames in commands for players LuckPerms has seen
- [x] Built-in contexts for world, gamemode, (and dimention 'type'?) - not currently implemented, but you can register context calculators from mods!
- [ ] Bugs? Might be! 🐞 

Enjoy :)

### DIFF
--- a/api/src/main/java/net/luckperms/api/platform/Platform.java
+++ b/api/src/main/java/net/luckperms/api/platform/Platform.java
@@ -78,7 +78,8 @@ public interface Platform {
         FABRIC("Fabric"),
         NEOFORGE("NeoForge"),
         FORGE("Forge"),
-        STANDALONE("Standalone");
+        STANDALONE("Standalone"),
+        HYTALE("Hytale");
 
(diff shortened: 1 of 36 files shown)


lucko:
@IamMusavaRibica thanks! I have just pushed a fix. Will release on CurseForge shortly

DeprecatedLuke:
#4220 to fix shutdown deadlock

PR #4244, opened 2026-03-27
Add Redis Sentinel support for messaging

ByteExceptionM:
Adds native Redis Sentinel support to the messaging system. Redis Sentinel provides high availability by monitoring master/replica instances and performing automatic failover when the master goes down - clients always get routed to the current master.

New `redis.sentinel` config section across all platforms with `enabled`, `master`, `addresses`, `username` and `password`. Sentinel credentials are separate from the main Redis credentials.

On the code side: five new config keys, a new `JedisSentineled` init path in `RedisMessenger` (with `parseAddress` extended for the sentinel default port 26379), sentinel-first check in `MessagingFactory`, and `isRedisAlive()` handling for the new type. Existing standalone and cluster setups are unaffected.

### DIFF
--- a/bukkit/src/main/resources/config.yml
+++ b/bukkit/src/main/resources/config.yml
@@ -268,6 +268,14 @@ redis:
   address: localhost
   username: ''
   password: ''
+  # Settings for Redis Sentinel.
+  # Sentinel provides high availability for Redis by monitoring master/replica instances.
+  # Port 26379 is used by default for sentinel nodes.
+  sentinel:
+    enabled: false
+    master: mymaster
+    addresses:
+      - localhost:26379
 
 # Settings for Nats.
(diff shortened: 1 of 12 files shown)


lucko:
Looks good! Sorry for the delay in reviewing.

Would it be possible to change it to reuse the existing `username`, `password` and `addresses` config properties?

ByteExceptionM:
Hey @lucko!

Reusing username/password won't work. Sentinel has its own auth, separate from the master. In real setups these are usually different credentials (Jedis even takes two separate configs for this).

Same with addresses: `redis.addresses` is the cluster node list, sentinel addresses are the watcher quorum.

I'd keep the `redis.sentinel.*` fields as they are.

lucko:
Thank you :)

PR #4248, opened 2026-04-03
Implement fabric-permission-api v1

lucko:


fabric/src/main/java/me/lucko/luckperms/fabric/listeners/FabricPermissionsApiV1Listener.java
```diff
@@ -0,0 +1,133 @@
...
+        this.plugin = plugin;
+    }
+
+    public void registerListeners() {
+        PermissionEvents.ON_REQUEST.register(this::onPermissionRequest);
+        PermissionEvents.PREPARE_OFFLINE_PLAYER.register(this::onPrepareOfflinePlayer);
+    }
+
+    private <T> T onPermissionRequest(PermissionContext ctx, PermissionNode<T> node) {
+        if (node.codec() == Codec.BOOL) {
```
lucko: @patbox You mentioned on the Fabric PR that this kind of Codec usage was incorrect, but I can't see another way to do it. Please could you elaborate on the intended usage?

Patbox: Checking for these common codecs is fine to handle them as a fast-path, but ideally you should also try parsing the value with it. In your case having it try to read the value as json and then parsing it with codec should be good enough (probably cached so to not have it reparsed from scratch every time)

lucko: Thanks, that makes sense, I have updated, does that look better?

Patbox: Yep looks good now

PR #4252, opened 2026-04-07
Add optional context filtering to group listmembers command

0ptk:
I used `ArgumentList#getContextOrEmpty` instead of `ArgumentList#getContextOrDefault` 
because the original behaviour of the command was to show all users/groups 
inheriting from the group regardless of context.


lucko:
Thanks!

PR #4255, opened 2026-04-15
Add index to time column in SQL Messenger

BrainStone:
Hey! I came across this while debugging some queries being run without an index and figured it was worth a quick fix.

This PR adds an index to the `time` column in the SQL Messenger table. While the polling (`SELECT`) is already efficient since it hits the primary key on `id`, the housekeeping (`DELETE`) currently requires a full table scan because it filters solely on `time`. 

Adding this index lets the database quickly locate and wipe those expired messages without scanning the whole table every 30 seconds. The overall impact isn't huge, but it's a "doesn't hurt" improvement that keeps things a bit cleaner and more efficient.

**Changes:**
- Added ``KEY (`time`)`` to the `CREATE TABLE` statement in `AbstractSqlMessenger`.
- Added an idempotent check/migration in `init()` to apply the index to existing tables.
- Verified this is safe for MySQL/MariaDB (the only supported engines for the `sql` messenger).

**Impact:**
- Fixes unindexed `DELETE` queries during housekeeping.
- Reduces database I/O and potential locking contention on the messenger table.

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/messaging/sql/AbstractSqlMessenger.java
+++ b/common/src/main/java/me/lucko/luckperms/common/messaging/sql/AbstractSqlMessenger.java
@@ -59,7 +59,7 @@ public abstract class AbstractSqlMessenger implements Messenger {
     public void init() throws SQLException {
         try (Connection c = getConnection()) {
             // init table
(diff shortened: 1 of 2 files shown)


lucko:
Thanks!

Issue #4256, opened 2026-04-16
Support SpongeAPI15+


SawFowl:
### Description

The plugin is currently not working. You need to add listening for the `ProvideServiceEvent` event to register the LuckPerms permissions service.

### Reproduction Steps

Install Sponge server with API 15 or later. Install LuckPerms and any other plugin that has any permissions, such as commands. All command permission checks will return `false` because the service is not registered.

### Expected Behaviour

Commands must be executed if the player has been granted permissions.

### Server Details

I tested on SpongeVanilla: 26.1.1-19.0.0-RC2587

### LuckPerms Version

LuckPerms v5.5.42

### Logs and Configs
```log
[22:56:32] [main/INFO] [plugin]: Loaded plugin(s): [minecraft, spongevanilla, sponge, spongeapi, worldedit, localeapi, luckperms, wasted, commandpack, guishopmanager, minerefill, clearlag, automessages, regionguard, chatmanager, protectionstones, tablistboard]
[22:56:32] [main/INFO] [c.sk8.wor.ext.pla.PlatformManager]: Got request to register class com.sk89q.worldedit.sponge.SpongePlatform with WorldEdit [com.sk89q.worldedit.extension.platform.PlatformManager@9da20f1]
[22:56:32] [main/INFO] [worldedit]: WorldEdit for Sponge (version 7.4.2+7450-eb8e82c) is loaded
[22:56:32] [main/INFO] [worldedit]: 
Plugin WorldEdit is using bStats Metrics  but currently has data sending disabled.
To change the enabled/disabled state of any bStats use in a plugin, visit the Sponge config!
[22:56:32] [main/INFO] [luckperms]: Loading configuration...
[22:56:32] [main/INFO] [luckperms]:         __    
[22:56:32] [main/INFO] [luckperms]:   |    |__)   LuckPerms v5.5.42
[22:56:32] [main/INFO] [luckperms]:   |___ |      Running on Sponge - SpongeVanilla
[22:56:32] [main/INFO] [luckperms]: 
[22:56:33] [main/INFO] [luckperms]: Loading storage provider... [HOCON]
[22:56:33] [main/INFO] [luckperms]: Loading internal permission managers...
[22:56:33] [main/WARN] [sponge]: No lookup found for listener me.lucko.luckperms.sponge.LPSpongePlugin$RegisterServiceListener.
[22:56:33] [main/INFO] [luckperms]: Performing initial data load...
[...]

PR #4257, opened 2026-04-17
Support SpongeAPI15+

SawFowl:
Now the plugin correctly registers its service on the Sponge server with API version 15 and higher. I personally tested on API 19 with game version 26.1.1. However, this will break compatibility with earlier versions due to API differences.
Fixed Issue https://github.com/LuckPerms/LuckPerms/issues/4256

### DIFF
--- a/sponge/build.gradle
+++ b/sponge/build.gradle
@@ -16,7 +16,7 @@ dependencies {
     implementation project(':sponge:sponge-service-proxy')
     compileOnly project(':common:loader-utils')
 
-    compileOnly('org.spongepowered:spongeapi:12.0.0') {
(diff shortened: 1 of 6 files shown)


lucko:
Sponge API 12 is still currently the LTS release, therefore that is what we are targeting. Is there a way to fix it whilst maintaining compatibility?

SawFowl:
I doubt that this is possible without reflection.

lucko:
Yes, in your commit, you have removed the reflection that was meant to handle it

SawFowl:
> Yes, in your commit, you have removed the reflection that was meant to handle it

No. Of course, you can return it, but the problem is registering the permissions service. There are quite serious differences between the registration methods for API12 and API15+.  I'm not even sure if it's even possible to support both variants at once.

PR #3615, opened 2023-04-01
Folia support

lucko:
Adds support for Paper's regionised/multithreading Folia project.

As you can see, there were fairly minimal changes required to get LuckPerms working, so I don't expect many bugs/issues. It has been lightly tested by me. :)

# Test build

There is a jar you can download from Jenkins: https://ci.lucko.me/job/LuckPerms-Folia/

PR #4263, opened 2026-05-11
Add Folia support for Bukkit module

JExcellence:
Detect Folia at runtime and route scheduler calls to the appropriate region-aware scheduler API. Uses reflection to maintain compile-time compatibility with the existing Paper API dependency.

All direct BukkitScheduler usages are replaced with Folia-compatible alternatives when running on a Folia server, while preserving the original behavior on standard Bukkit/Spigot/Paper servers.


lucko:
What's wrong with #3615?

JExcellence:
Nothing, I just read your first comment with the link. My Bad! Glad it's being pushed, ty. 

I have overseen this.

PR #4271, opened 2026-05-31
Fix NeoForge command source permission checks

ustc-zzzz:
This PR depends on neoforged/NeoForge#3154, which introduced `CommandSourceStack#getSourceProfile` and solved the platform-side issue where the actual command source profile was not available from `CommandSourceStack`. Without it, mods had to infer the player from `getEntity()`, which is incorrect for commands like `execute as Bob run ...`.

This PR updates NeoForge permission checks to use the source profile instead of the command entity. This prevents command permissions from being checked against the target entity selected by `execute as`, rather than the actual command executor.


common/minecraft/src/main/java/me/lucko/luckperms/common/minecraft/command/BrigadierInjector.java
```diff
@@ -140,19 +137,10 @@ private InjectedPermissionRequirement(MinecraftLuckPermsPlugin<?, ?> plugin, Str
 
         @Override
         public boolean test(CommandSourceStack source) {
-            if (source.getEntity() instanceof ServerPlayer player) {
+            Tristate state = this.plugin.getSenderFactory().wrap(source).getPermissionValue(this.permission);
```
lucko, line 140: The sender factory is only really used by internal LP commands and isn't intended to be used for permission functionality, could this be refactored with that in mind?

PR #4273, opened 2026-06-03
Fix: Execute finishCurrentTask on Server Thread

### LINKED ISSUE #4272: In AsyncConfigurationTask, `finishCurrentTask` should be invoked on Server Thread
### Description

In AsyncConfigurationTask, `finishCurrentTask` should be invoked on Server Thread

However, the codes here invoked it on CompletableFuture's shared workers.

https://github.com/LuckPerms/LuckPerms/blob/master/neoforge/src/main/java/me/lucko/luckperms/neoforge/util/AsyncConfigurationTask.java#L56

### Reproduction Steps

/

### Expected Behaviour

A MinecraftServer object should be passed to `whenCompleteAsync` as the specific executor.

### Server Details

NeoForge, 26.1.2.68-beta

### LuckPerms Version

5.5-SNAPSHOT

### Logs and Configs

Sorry I only got an screen shot.

Image

### Extra Details

Image

> Hi, I'm wondering whether finishCurrentTask should be invoked on threads except 'Server Thread'
> No

burningtnt:
Fix #4272


lucko:
A slightly different fix, but should have the same effect :) cd0f009e0fdde1002a32c8f3c3fe0818d26a88e4

PR #4279, opened 2026-06-30
Add Waterdog platform support to LuckPerms

SenseiJu:


stevei5mc:
An error occurs when using YAML as the storage method.

[server.log](https://github.com/user-attachments/files/30611609/server.log)

PR #4280, opened 2026-07-04
Add MongoDB Indexes

### LINKED ISSUE #4246: add MongoDB indexes
### Description

Luckperms doesnt create Indexes in MongoDB

Image

For example: luckperms_uuid collection without name index can be slow (I have 32k+ accounts).
When I create the "name" index, fixed.

I migrated my database from MySQL, I don't know if that was the reason.

### Proposed Behaviour

Create Indexes in MongoDB code.

### Extra Details

_No response_

masonakdev:
- Added an index on `uuid.name` to match the SQL `players.username` index.
- Index is created automatically and idempotently during DB init.
- Added an integration test.
- Tested with `./gradlew :common:test -PdockerTests --tests me.lucko.luckperms.common.storage.MongoStorageTest`.

This is my first PR for this project. Will tackle a bugfix next but this seemed like a pretty easy kill. If there's anything wrong with style or anything else, just let me know. Used to use LuckPerms a lot back in the day, really cool project.

Closes #4246.

### DIFF
--- a/common/src/test/java/me/lucko/luckperms/common/storage/MongoStorageTest.java
+++ b/common/src/test/java/me/lucko/luckperms/common/storage/MongoStorageTest.java
@@ -25,37 +25,116 @@
 
 package me.lucko.luckperms.common.storage;
 
+import com.mongodb.MongoClient;
+import com.mongodb.ServerAddress;
+import com.mongodb.client.MongoDatabase;
 import me.lucko.luckperms.common.plugin.LuckPermsPlugin;
 import me.lucko.luckperms.common.storage.implementation.StorageImplementation;
 import me.lucko.luckperms.common.storage.implementation.mongodb.MongoStorage;
 import me.lucko.luckperms.common.storage.misc.StorageCredentials;
+import org.bson.Document;
 import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.utility.DockerImageName;
+
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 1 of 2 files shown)


lucko:
Thanks, looks good!

While we are adding indexes, it is worth adding more than just that one? Have a look at the mysql sql file - there are three in total. (or better may be worth analysing the MongoStorage class to see how we do queries)

common/src/test/java/me/lucko/luckperms/common/storage/MongoStorageTest.java
```diff
@@ -25,37 +25,84 @@
...
+            assertEquals(
+                    new Document("name", 1),
+                    nameIndex.get("key", Document.class)
+            );
+            assertEquals(
+                    1L,
+                    indexes.stream()
+                            .filter(index -> "name_1".equals(index.getString("name")))
+                            .count()
+            );
```
lucko, line 111: assume this is checking that it hasn't been duplicated? could you add a short comment to that effect

masonakdev: Absolutely, that was a bit unclear.

masonakdev:
> Thanks, looks good!
> 
> While we are adding indexes, it is worth adding more than just that one? Have a look at the mysql sql file - there are three in total. (or better may be worth analysing the MongoStorage class to see how we do queries)

Great point, bit sloppy on my part. I believe I've added the indexes that should exist and also refactored/cleaned up the integration test.

I hate to ask for hand-holding, but if there are more indexes that would be worth adding would you mind pointing me in a direction?

Edit: Amending commit message and PR title to match changes.

PR #4283, opened 2026-07-14
Fix `NodeMapMutable#setContent` exposing a partially-populated map to concurrent readers

ByteExceptionM:
`setContent` does `clear()` then `addAll()`. Reads are lock-free, so one landing mid-operation sees an empty/half-filled map. It runs on every messaging sync, and since storage + messaging share a worker pool, a save reading `asList()` can overlap a reload on the same holder. Full-snapshot backends (`Mongo`/flatfile) then persist the truncated state and propagate it fleet-wide. Hit 3x in prod: `default` (~400 nodes) dropped to ~10% network-wide.

Fix: build new content in a detached `NodeMapMutable`, swap maps in only once fully populated, so readers always see complete-old or complete-new. Map fields made `volatile`. `Difference` kept identical by recording old nodes as `REMOVE` before the swap (reproduces old `clear()`+`addAll()`); inverse pairs cancel, so only real changes hit the delta.

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/model/nodemap/NodeMapMutable.java
+++ b/common/src/main/java/me/lucko/luckperms/common/model/nodemap/NodeMapMutable.java
@@ -72,13 +72,14 @@ public class NodeMapMutable extends NodeMapBase {
      *
      * We use our own "multimap"-like implementation here because guava's is not thread safe.
      *
-     * The map fields aren't final because they are replaced when large updates (e.g. clear)
-     * are performed. We do this so there's no risk that the read methods will see an inconsistent
-     * state in the middle of an update from the DB. (see below comment about locking - we don't
-     * lock for reads!)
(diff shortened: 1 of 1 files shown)


lucko:
Slightly different approach proposed in the PR linked above.

I think the bug is still there with either fix - which is that Mongo storage (unlike flat files which use locks and SQL which uses the Difference/exportChanges system) can concurrently load user/group data, replacing the whole map, and save user/group data, persisting the current state of the map. This design flaw unfortunately enables changes to be lost/overwritten. The fixes proposed just make it less noticeable, as the update churn is significantly reduced.

I think MongoStorage probably needs a wider re-think, probably updated to use the Difference/exportChanges system, the same as the SQL storage impl.

lucko:
The other two PRs linked above have been merged, so going to close this :) thanks for the detailed report/PR, let me know if you notice any issues :)

Issue #4211, opened 2026-01-11
NeoForge Luckperms v5.4.150


rackgaming:
### Description

`[01:23:50] [luckperms-command-executor/WARN] [luckperms/]: An exception was thrown by me.lucko.luckperms.neoforge.context.NeoForgePlayerCalculator whilst estimating potential contexts
java.lang.ClassCastException: class net.minecraft.core.MappedRegistry$1 cannot be cast to class net.minecraft.core.Registry (net.minecraft.core.MappedRegistry$1 and net.minecraft.core.Registry are in module minecraft@1.21.1 of loader 'TRANSFORMER' @3fcee3d9)
        at java.base/java.util.Optional.ifPresent(Optional.java:178) ~[?:?] {re:mixin}
        at me.lucko.luckperms.neoforge.context.NeoForgePlayerCalculator.estimatePotentialContexts(NeoForgePlayerCalculator.java:103) ~[?:?] {}
        at me.lucko.luckperms.common.context.manager.ContextManager.getPotentialContexts(ContextManager.java:181) ~[?:?] {}
        at me.lucko.luckperms.common.webeditor.WebEditorRequest.generate(WebEditorRequest.java:133) ~[?:?] {}
        at me.lucko.luckperms.common.webeditor.WebEditorSession.create(WebEditorSession.java:60) ~[?:?] {}
        at me.lucko.luckperms.common.commands.misc.EditorCommand.execute(EditorCommand.java:110) ~[?:?] {}
        at me.lucko.luckperms.common.command.abstraction.SingleCommand.execute(SingleCommand.java:55) ~[?:?] {}
        at me.lucko.luckperms.common.command.abstraction.SingleCommand.execute(SingleCommand.java:47) ~[?:?] {}
        at me.lucko.luckperms.common.command.CommandManager.execute(CommandManager.java:280) ~[?:?] {}
        at me.lucko.luckperms.common.command.CommandManager.lambda$executeCommand$1(CommandManager.java:184) ~[?:?] {}
        at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1804) ~[?:?] {}
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144) ~[?:?] {}
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642) ~[?:?] {}
[...]

PR #4294, opened 2026-08-04
fix(context): prevent ClassCastException when estimating dimension-type contexts

BlockException:
Some mod loader registry implementations (observed on **NeoForge 1.21.1**)
can return a registry from **RegistryAccess#lookup(...)** that isn't
safely castable at this call site (e.g. MappedRegistry$1). Catch the
**ClassCastException** so the whole potential-context estimation (used by
the web editor) doesn't fail.

**Fixes** #4211


lucko:
I would argue it's a NeoForge/API issue - I would not expect that code to throw the exception it does :(

Regardless of that, I appreciate you fixing and opening a PR :)

Issue #4292, opened 2026-08-04
Messages are broken on Bukkit impl


Outfluencer:
### Description

Message output is broken because of the outdated adventure-bukkit lib (your fork is based on 4.4.0)

Clicking on permission nodes of a group would paste the according delete command into the clients chat.
This works fine on BungeeCord but not on Bukkit.

### Reproduction Steps

/lp group default permission set test

/lp group default permission info

Click on the command feedback permission node.

### Expected Behaviour

The  permission nodes should be clickable and put the delete command in the clients chatbox

### Server Details

paper 1.21.11 latest

### LuckPerms Version

5.5.65

### Extra Details

I have fixed it locally by depending on net.kyori.adventure-platform-bukkit and changing the     ADVENTURE_PLATFORM_BUKKIT enum in the Dependency.java class of Bukkit to

```
ADVENTURE_PLATFORM_BUKKIT(
        "net{}kyori",
        "adventure-platform-bukkit",
        "4.4.1",
        "zcs6Zqu/njk2FNtd6yf4QRF8W1t+9SdToshG/UbJzfE=",
        Relocation.of("adventure", "net{}kyori{}adventure"),
        Relocation.of("adventure.util.option", "net{}kyori{}option")
    )
```

lucko:
Fixed by above commit - adventure-platform is unsupported so using the native Paper impl is a better long term fix I think. :)

Outfluencer:
> Fixed by above commit - adventure-platform is unsupported so using the native Paper impl is a better long term fix I think. :)

Yes it is, thank you!

PR #4297, opened 2026-08-11
Fix inverted Audience check in PaperAdventureBridge

AndyOctopus:
The guard added in #4292 tests the wrong direction, so it never fires:

```java
if (CommandSender.class.isAssignableFrom(audienceClass)) {
    throw new IllegalStateException("CommandSender does not implement Audience");
}
```

`A.isAssignableFrom(B)` is true when `B` is a subtype of `A`, so this asks whether `Audience` is a subtype of `CommandSender`. That is never true, and the message states the opposite of the condition being tested. The check that is meant to reject platforms without native Adventure support therefore always passes.

### Why that matters

`Class.forName("net.kyori.adventure.audience.Audience")` succeeding does not imply the server's `CommandSender` implements it. LuckPerms relocates its own copy of Adventure (which is why the lookup string is split), but on Bukkit the plugin classloader also resolves classes loaded by *other* plugins, so any plugin on the server that ships Adventure unrelocated makes all three `Class.forName` calls succeed.

With the guard inoperative, `PaperAdventureBridge.INSTANCE` is non-null, `BukkitSenderFactory.create` picks `PaperBridgeBukkitSenderFactory`, and the first message sent is the startup banner in `AbstractLuckPermsPlugin#enable`:

```
Error occurred while enabling LuckPerms v5.5.71
java.lang.IllegalArgumentException: object is not an instance of declaring class
    at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(...)
    at me.lucko.luckperms.bukkit.util.PaperAdventureBridge.sendMessage(PaperAdventureBridge.java:78)
    at me.lucko.luckperms.bukkit.BukkitSenderFactory$PaperBridgeBukkitSenderFactory.sendMessage0(BukkitSenderFactory.java:129)
    ...
    at me.lucko.luckperms.common.plugin.AbstractLuckPermsPlugin.enable(AbstractLuckPermsPlugin.java:158)
```

`IllegalArgumentException` is unchecked, so `catch (ReflectiveOperationException)` in `sendMessage` does not apply and it propagates out of `enable()` — the plugin fails to enable entirely rather than falling back to adventure-platform.

Observed on a 1.8.8 server running 5.5.71, but nothing about the failure is version specific: it needs only a `CommandSender` that does not implement `Audience` plus any plugin exposing unrelocated `net.kyori.adventure`.

### The change

```java
if (!audienceClass.isAssignableFrom(CommandSender.class)) {
    throw new IllegalStateException("CommandSender does not implement Audience");
}
```

On Paper this is unchanged behaviour: `CommandSender extends Audience`, and `PluginClassLoader` delegates to the parent first, so `audienceClass` is the server's own `Audience`. Where the assertion does not hold, the constructor now throws as it was meant to and `create` falls back to `AdventurePlatformBukkitSenderFactory`.

I have kept this to the one line. Happy to follow up separately with a change that prevents an unexpected runtime exception from the bridge failing enable outright, but that felt like a design call rather than part of this fix.

### DIFF
--- a/bukkit/src/main/java/me/lucko/luckperms/bukkit/util/PaperAdventureBridge.java
+++ b/bukkit/src/main/java/me/lucko/luckperms/bukkit/util/PaperAdventureBridge.java
@@ -55,7 +55,7 @@ public final class PaperAdventureBridge {
         Class<?> componentClass = Class.forName(adventurePkg + "text.Component");
         Class<?> serializerClass = Class.forName(adventurePkg + "text.serializer.gson.GsonComponentSerializer");
 
-        if (CommandSender.class.isAssignableFrom(audienceClass)) {
+        if (!audienceClass.isAssignableFrom(CommandSender.class)) {
             throw new IllegalStateException("CommandSender does not implement Audience");
         }
 


lucko:
Thanks :)

PR #4302, opened 2026-08-15
Prevent NPE in PostgresMessenger when SocketException message is null

Smorki:
`e.getCause().getMessage()` can return `null` (as `Throwable#getMessage` is nullable). Calling `.equals(...)` on it throws a `NullPointerException`, which prevents the intended suppression of the "Socket closed" case.

Swapped the operands so the string literal is on the left side. `"Socket closed".equals(null)` safely returns `false` instead of throwing.

### DIFF
--- a/common/src/main/java/me/lucko/luckperms/common/messaging/postgres/PostgresMessenger.java
+++ b/common/src/main/java/me/lucko/luckperms/common/messaging/postgres/PostgresMessenger.java
@@ -150,7 +150,7 @@ public class PostgresMessenger implements Messenger {
                 }
 
             } catch (PSQLException e) {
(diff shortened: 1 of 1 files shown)


lucko:
Hey, thanks for the PR - have you observed this to be the case (ie seen the error in a log on a live server), or is this change just preventative?

Smorki:
Hey, no - this is preventative I noticed it while reading through the code since Throwable.getMessage() is nullable Happy to close if you'd rather only merge fixes for observed issues

lucko:
> Happy to close if you'd rather only merge fixes for observed issues

observed or at least reproducible, yes - thanks :)

PR #4310, opened 2026-08-26
Replace Identifiers path separator with Luckperms separator

DrexHD:
Fabric permission api v1 recommends permission id's like this: `Identifier.fromNamespaceAndPath("mymod", "command/main")`. These paths don't work well with luckperms permission id system and should therefor use luckperms `.` path separator instead.


lucko:
Inclined to agree given we already replace the `:` - thanks!

PR #4311, opened 2026-08-26
chore: update to latest hytale version

oskarscot:
Updates the hytale module to the latest pre-release which will be the same version as tomorrow's Update 6 release. There will be no api changes tomorrow on release vs what is current published under 0.6.0-pre.13.1


lucko:
Thanks for the PR - sorry I meant to comment just after my commit. As you can see, slightly different approach to supporting `getUsersWithPermission` but more or less the same otherwise :)

Thanks again!

Issue #4268, opened 2026-05-22
User data for connecting player is housekept if player spends too long in the configuration stage


Leymooo:
### Description

If player stuck for too long inside Configuration state (1.21.2+), cache will forget user and then on PlayerJoinEvent user is not loaded, so plugins are failing to correctly check permissions or do some stuff (TAB/WorldGuard/Essentials). 

```
[18:28:38] [User Authenticator #279/INFO]: UUID of player MEGAPbICHAR is aca96453-2491-3113-aae9-dd83004c6b50
[18:31:20] [TAB Processing Thread/INFO]: [TAB] [WARN] Player MEGAPbICHAR's group (none) is not in sorting list! Sorting list: герцог,князь,принц,титан,элита,глава,сквид,аспид,герой,страж,барон,default. Player will be sorted on the bottom.
[18:31:20] [Server thread/ERROR]: Could not pass event ProcessPlayerEvent to WorldGuard v7.0.17-SNAPSHOT+89344c4
com.google.common.util.concurrent.UncheckedExecutionException: me.lucko.luckperms.bukkit.vault.ServerThreadLookupException: A Vault API request has been made on the main server thread that LuckPerms cannot safely respond to.
This is NOT a bug - please do not report it to LuckPerms.
Instead, please carefully read the information given below.
[18:31:20] [Server thread/ERROR]: Could not pass event ProcessPlayerEvent to WorldGuard v7.0.17-SNAPSHOT+89344c4
com.google.common.util.concurrent.UncheckedExecutionException: me.lucko.luckperms.bukkit.vault.ServerThreadLookupException: A Vault API request has been made on the main server thread that LuckPerms cannot safely respond to.
This is NOT a bug - please do not report it to LuckPerms.
Instead, please carefully read the information given below.

Full here: https://mclo.gs/oy89ydQ

[18:31:20] [Server thread/INFO]: MEGAPbICHAR joined the game
[18:31:20] [Server thread/INFO]: MEGAPbICHAR[/82.209.93.56:59991] logged in with entity id 7664060 at ([lobby]-265.94941383923594, 82.0, 38.607604489965524)
[18:31:20] [luckperms-command-executor/INFO]: [LP] Set ftutils.cooldown.clientversionwarning to true for megapbichar for a duration of 1 hour in context global.
```

Notice time deltas between AsyncPreLoginEvent & PlayerJoinEvent

### Reproduction Steps

[...]

PR #4313, opened 2026-09-09
fix: keep user data loaded while player is in configuration phase

HEBEI77:
See #4268.

Problem: a player stuck in the configuration phase for 2+ minutes (resource pack download) has their user unloaded by the housekeeper. On PlayerJoinEvent the user is not loaded, so permission checks fail.

Impact: Vault requests on join throw ServerThreadLookupException, groups resolve wrong.

Root cause: the recentlyUsed window is 2 minutes and LPBukkitBootstrap.isPlayerOnline uses Server.getPlayer, which returns null until configuration ends. Both guards miss, cleanup unloads.

Fix: track configuring UUIDs from PlayerConnectionValidateLoginEvent in a 15 minute expiring set, remove on join and quit, and count tracked UUIDs as online in isPlayerOnline. Registration is gated on the Paper event class existing, Spigot and old versions are untouched. Untracked offline players behave as before.

Test: verified every Paper signature used against the folia-api jar this module compiles with. No build or in-game run here.


lucko:
Thanks for the PR :)

I have implemented using a slightly different approach so it works for all platforms, not just Paper: https://github.com/LuckPerms/LuckPerms/commit/24a25458d66e2fc976a86f2633f8c62956c821a5

Issue #4312, opened 2026-09-05
Recent kotlin fails to compile because of missing transitive annotations


brickmonster:
### Description

Kotlin 2.4.0+ fails to compile because luckperms api does not include checker-qual transitively. It was a warning in earlier versions.

```
src/main/kotlin/Main.kt:2:56 Type annotation class 'org.checkerframework.checker.nullness.qual.NonNull' of the inferred type is inaccessible. Check the module classpath for missing or conflicting dependencies. This will become an error in language version 2.4. See https://youtrack.jetbrains.com/issue/KT-80247.
```

### Reproduction Steps

[qual-repro.zip](https://github.com/user-attachments/files/31871144/qual-repro.zip)

### Expected Behaviour

It compiles lol 

### Server Details

N/A

### LuckPerms Version

N/A

lucko:
Should be fixed in https://github.com/LuckPerms/LuckPerms/commit/e794facb05051f4ef18f4176b73e28fa0b87291d

PR #4314, opened 2026-09-12
Publish checker-qual for API consumers

HP-network:
Kotlin 2.4 treats the missing checker-qual annotation class as a compilation error when consuming the API artifact.\n\nThe API exposes Checker Framework annotations in its public signatures, but the dependency is currently compile-only and absent from the published metadata. This changes it to a compile-only API dependency so Maven and Gradle consumers receive it transitively without bundling it into the API jar.\n\nVerified with the reproduction from #4312:\n- baseline consumer: fails with NonNull annotation class inaccessible\n- modified consumer: compiles successfully\n\nFixes #4312


lucko:
Slightly different approach: https://github.com/LuckPerms/LuckPerms/commit/e794facb05051f4ef18f4176b73e28fa0b87291d

Thanks for the PR :)

Issue #3958, opened 2024-08-16
FeedbackForwardingSender can't run commands


masmc05:
### Description

If a FeedbackForwardingSender is created using [Bukkit#createCommandSender](https://jd.papermc.io/paper/1.21.1/org/bukkit/Bukkit.html#createCommandSender(java.util.function.Consumer)) it doesn't have permission to run luckperms commands

### Reproduction Steps

Create a simple command sender and dispatch a luckperms command with it

### Expected Behaviour

For the command specified by the user (for example in a config) to be able to run with the forwarding sender

### Server Details

1.21.1-16-master@bf5852a (2024-08-16T07:29:16Z)

### LuckPerms Version

LuckPerms version 5.4.139

Laarryy:
Looks like a legitimate bug, but not certain it's with LP. Does your CommandSender succeed in sending other operator-level commands? If so, we'll take a closer look, and if you would like to PR a resolution please feel free.

lucko:
LP has special behaviour for `ConsoleCommandSender` and `RemoteConsoleCommandSender` or subclasses of those. (https://github.com/LuckPerms/LuckPerms/blob/master/bukkit/src/main/java/me/lucko/luckperms/bukkit/BukkitSenderFactory.java#L97)

Otherwise, it just defers to `CommandSender.hasPermission`, so I suspect that is returning false.

PR #4315, opened 2026-09-15
Treat Paper's FeedbackForwardingSender as a console sender

ThoriaDevelopment:
A `CommandSender` created with `Bukkit#createCommandSender` (Paper's `io.papermc.paper.commands.FeedbackForwardingSender`) is console-equivalent, but LuckPerms' `isConsole` check does not recognize it, so everything keyed off that classification treats it as an ordinary sender.

Paper's javadoc for `createCommandSender` says the returned sender has "the same effective permissions as `getConsoleSender()`", and the implementation passes the console's permission set into the sender's superclass, so `hasPermission` answers exactly as the console does. `BukkitSenderFactory#isConsole` covered `ConsoleCommandSender` and `RemoteConsoleCommandSender` only, which left the third console sender on the wrong side of the check.

`BukkitSenderFactory#isConsole` now also recognizes Paper's `FeedbackForwardingSender` by class name. The comparison does nothing on Spigot, where the class does not exist, and keeps working on Paper forks that inherit it. The class is final, so the name comparison is exact. I avoided keying off `isOp()`, because command blocks report `isOp() == true` and must not gain console permissions.

Testing was done on throwaway local Paper servers with only LuckPerms and a probe plugin that calls `Bukkit#createCommandSender` and dispatches commands through the returned sender. With `disable-luckperms-commands.players: true` and `console: false` on Paper 26.2 build 123:

- master: dispatching `lp user <name> info` through the sender replies "LuckPerms commands are disabled."
- this branch: the same dispatch reaches argument resolution, and the reply is forwarded to the consumer.

With the default config, command authorization also succeeds on master, because the sender inherits the console's permission set and `hasPermission` returned true for every node I tried, on Paper 1.21.1 build 133 and Paper 26.2 build 123. This change is therefore about which of the two `disable-luckperms-commands` settings applies to the sender, and about `isConsole()` reporting it as console to the rest of the plugin, not about a permission failure.

Not tested: Spigot and the other platform modules. The change is confined to the Bukkit sender factory and refers to no Paper types, so it stays inert where the class is absent.

Build: `./gradlew build -x :standalone:test` passes on Temurin JDK 25. `:standalone:test` is excluded because its flat-file export tests compare generated LF output against expected resources that `core.autocrlf` checked out as CRLF on Windows. The same six `StorageIntegrationTest` cases fail on master without this change.

Refs #3958.


lucko:
Should we treat it as the console though?

What's your use case? or are you just fixing it because you saw the issue?

ThoriaDevelopment:
I picked this up from #3958 rather than from a server of my own, so there is no production use case behind it. That turned out to matter, because when I tested it properly I could not reproduce the failure the issue reports.

What I ran, on throwaway local servers with only LuckPerms and a small probe plugin that calls `Bukkit#createCommandSender` and dispatches commands through the returned sender:

- Paper 1.21.1 build 133 (Java 21) and Paper 26.2 build 123 (Java 25), each with LuckPerms built from master and from this branch.
- On both, the sender is `io.papermc.paper.commands.FeedbackForwardingSender`, `isOp()` returns true, and `hasPermission` returns true for every node I tried, including `luckperms.user.info`, `luckperms.creategroup`, and two nodes my probe plugin defines with op and true defaults. The console sender returns the same answers. Paper's `createCommandSender` javadoc says the sender has "the same effective permissions as `getConsoleSender()`", and the implementation passes the console's permission set into the sender's superclass, so this holds by design.
- Because of that, master already runs the commands through this sender: `lp user ProbeUser info` reached argument resolution ("A user for ProbeUser could not be found") and `lp creategroup probe_probe` created the group, with both replies forwarded to the consumer. This branch produces the same results.

So the permission claim in my description was wrong, and I have replaced it. The only difference I could reproduce between master and this branch is which of the two config settings applies to the sender. With `disable-luckperms-commands.players: true` and `console: false`, master replies to the forwarding sender with "LuckPerms commands are disabled.", while this branch lets the command through as a console sender.

That is a much smaller change than the description claims, and it is a judgement call about classification rather than a fix for a permission failure. If you would rather not carry a Paper class-name check in `isConsole` for it, say so and I will close the PR. If you do want the sender classified as console, I will leave the diff as it is and answer in the description instead of the issue.

